PR #22, opened 2023-07-24
Fix market resolution date calculation

angrybayblade:


Adamantios (approved):

README.md
```diff
@@ -44,7 +44,8 @@ Market maker has several configurable parameters for creating markets. These par
 - `num_markets`: Number of markets an agent to allowed to create before, default is 1
 - `market_fee`: Fees for creating a market, default is 2 unit (Eth, xDAI, etc...)
 - `initial_funds`: Initial funds for the market, default is 1 unit (Eth, xDAI, WxDAI, etc...)
-- `market_timeout`: Time for which the market should be active after opening, default is 7 days
+- `market_timeout`: Maximum time allowed for finalizing answer on `realitio`, default is 7 days
```
jmoreira-valory, line 47: Why 7 days? I think that's a lot. It should be 1 day.

packages/valory/skills/market_creation_manager_abci/behaviours.py
```diff
@@ -290,19 +290,22 @@ def _get_llm_response(self) -> Generator[None, None, Optional[dict]]:
...
             try:
+                # Date of the outcome
                 resolution_date = parse_date_timestring(q["resolution_date"])
                 if resolution_date is None:
                     self.context.logger.error(
                         "Cannot parse datestring " + q["resolution_date"]
                     )
                     continue
-                if resolution_date < minimum_resolution_date:
+                if resolution_date < minimum_opening_date:
```
jmoreira-valory, line 308: Is this OK??

resolution_date = Date that the event happens ("occurring date")
minimum_opening_date= Opening date on Realitio (i.e., the event should be known when opening the question on REalitio)

Therefore seems the check is not OK,  no??

jmoreira-valory: Another question, what is the date the Omen market is opened? The condition should ideally be:

omen_open_date < resolution_date (ocurring date) < realitio_open_date (= omen_closing_date)

README.md
angrybayblade: It's configurable

packages/valory/skills/market_creation_manager_abci/behaviours.py
angrybayblade: 1. The check is correct. Yes, the `resolution_date` refers to the date when the event will be occurring. Yes, `minimum_opening_date ` refers to the opening date on omen, but it's not something that is statically defined. We calculate this value based on the current date and the [minimum_market_time](https://github.com/valory-xyz/market-creator/pull/22/files/7a2e1bcf4fdeadbc900085e17c1a73a7c638ccd2#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R48) parameter. What this means is the minimum opening date needs to be at least `current_date + minimum_market_time` or greater. 

3. The same date when the tx gets executed

README.md
jmoreira-valory: Please change this to 1 day by default in all referenced components and env variables.

jmoreira-valory (approved):

PR #30, opened 2023-07-26
Fix market creation counter

angrybayblade:


packages/valory/skills/market_validation_abci/rounds.py
```diff
@@ -78,14 +78,12 @@ def end_block(self) -> Optional[Tuple[BaseSynchronizedData, Event]]:
...
-                SynchronizedData, self.synchronized_data
-            ).markets_created
-            markets_created += 1
             synchronized_data = self.synchronized_data.update(
                 synchronized_data_class=SynchronizedData,
                 **{
-                    get_name(SynchronizedData.markets_created): markets_created,
+                    get_name(SynchronizedData.markets_created): cast(
+                        SynchronizedData, self.synchronized_data
+                    ).markets_created,
```
angrybayblade, line 86: Before we used to increment the counter after the market was actually created, IE after the tx settlement round

0xArdi (approved):

PR #37, opened 2023-08-01
Update readme

jmoreira-valory:
Update readme


README.md
```diff
@@ -19,117 +22,163 @@ Market creator (market maker) is an autonomous service created with the [Open Au
...
-```
+## Configure the service
 
-Now running an agent this way, will not create any market since the agent depends on gnosis multisig contract to execute the final transactions which actually creates the agent. So if you want to run an agent end 2 end you will require a gnosis multisig safe with 1 registered member. To do this run following in your terminal
+Set up the following environment variables, which will modify the performance of the trading agent. **Please read their description below**. We provide some defaults, but feel free to experiment with different values. Note that you need to provide `YOUR_AGENT_ADDRESS` and `YOUR_SAFE_ADDRESS` from the section above.
 
 ```bash
-aea generate-key ethereum
+export ETHEREUM_LEDGER_RPC=INSERT_YOUR_RPC
+export ETHEREUM_LEDGER_CHAIN_ID=100
```
dagacha, line 85: Not for now, but we need to change the variable name `ETHEREUM_LEDGER_RPC` to `RPC` or similar, but it implies changes in the code.

dagacha (approved):
LGTM

README.md
jmoreira-valory: I discussed this with Ardian. In this case `ETHEREUM` refers to `ETHEREUM`-type ledger (instead of cosmos), not `ETHEREUM` chain. But I agree this needs to be changed.

PR #38, opened 2023-08-01
feat: add deposit round

0xArdi:
This PR adds support for depositing xDAI.


packages/valory/skills/market_creation_manager_abci/rounds.py
```diff
@@ -487,35 +514,59 @@ class MarketCreationManagerAbciApp(AbciApp[Event]):
...
             Event.ERROR: CollectRandomnessRound,
-            Event.NO_MARKETS_RETRIEVED: SkippedMarketCreationManagerRound,
+            Event.NO_MARKETS_RETRIEVED: DepositDaiRound,
         },
         PrepareTransactionRound: {
             Event.DONE: FinishedMarketCreationManagerRound,
             Event.NO_MAJORITY: CollectRandomnessRound,
             Event.ROUND_TIMEOUT: CollectRandomnessRound,
         },
+        DepositDaiRound: {
```
jmoreira-valory, line 524: Doesn't it make more sense this:

RetrieveApprovedMarketRound --done--> Deposit --> PrepareTransaction
RetrieveApprovedMarketRound --no_markets--> SyncMarketsRound ?

Otherwise, if we have DAI but not WDAI, it will fail the transaction...

0xArdi: we cannot transition from deposit to prepare transaction, however I see your point, we can move the deposit tx before the market creation stuff

PR #42, opened 2023-08-02
Fixes data collection for market liquidity removal by using subgraph instead of filtering `on-chain` events 

angrybayblade:


packages/valory/skills/market_creation_manager_abci/behaviours.py
```diff
@@ -474,51 +506,36 @@ def _get_markets_with_funds(
...
-        if contract_api_response.performative != ContractApiMessage.Performative.STATE:
-            self.context.logger.error(
-                f"Failed to get condition preparation events: {contract_api_response}"
-            )
-            return None
-        condition_preparation_events = cast(
-            Optional[List[Dict[str, Any]]],
-            contract_api_response.state.body.get("data", []),
+        query = CONDITIONS_QUERY.substitute(conditions=json.dumps(condition_ids))
+        response = yield from self.get_http_response(
```
angrybayblade, line 513: We will have to add a retry mechanism here, but first let's test if this works or not

packages/valory/skills/market_creation_manager_abci/behaviours.py
```diff
@@ -474,51 +506,36 @@ def _get_markets_with_funds(
...
-        )
-        if contract_api_response.performative != ContractApiMessage.Performative.STATE:
-            self.context.logger.error(
-                f"Failed to get question: {contract_api_response}"
-            )
-            return None
-        questions = cast(
-            Optional[List[Dict[str, Any]]],
-            contract_api_response.state.body.get("data", []),
+        query = QUESTIONS_QUERY.substitute(conditions=json.dumps(question_ids))
```
angrybayblade, line 529: Same as above

Adamantios (approved):

PR #57, opened 2023-08-24
Fix failed liquidity transactions

jmoreira-valory:
Fix failed liquidity transactions.

This error is caused by this: https://github.com/ethereum/eth-abi/blob/master/docs/release_notes.rst#breaking-changes-1

### DIFF
--- a/packages/valory/contracts/fpmm/contract.py
+++ b/packages/valory/contracts/fpmm/contract.py
@@ -180,7 +180,7 @@ class FPMMContract(Contract):
             )
 
             # Encode the input data (constructor params)
-            encoded_input_data = ledger_api.api.codec.encode_abi(
+            encoded_input_data = ledger_api.api.codec.encode(
                 ["address[]", "address"], [markets, safe_address]
             )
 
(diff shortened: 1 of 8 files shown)


packages/valory/contracts/fpmm/contract.py
```diff
@@ -171,31 +171,37 @@ def get_markets_with_funds(
...
+            # the decoding returns a Tuple with a single element so we need to access the first element of the tuple,
+            # which contains a tuple of markets that have funds
+            non_zero_markets = [
+                ledger_api.api.to_checksum_address(market_address)
+                for market_address in ledger_api.api.codec.decode(["address[]"], encoded_markets)[0]
+            ]
+
+            _logger.info(f"Markets with non-zero funds retrieved: {non_zero_markets}")
+        except Exception as e:
+            _logger.error("An exception occurred in get_markets_with_funds():", str(e))
```
DavidMinarsch, line 205: If we keep this in then we will not get an "Error" performative when there's an error... is that what you want?

0xArdi (approved):

packages/valory/contracts/fpmm/contract.py
jmoreira-valory: Fixed

DavidMinarsch (approved):

PR #65, opened 2023-10-11
Feat/market approval update

0xArdi:
This PR adds support for marking markets as created on the market approval server.

### DIFF
--- a/packages/valory/skills/market_creation_manager_abci/behaviours.py
+++ b/packages/valory/skills/market_creation_manager_abci/behaviours.py
@@ -80,8 +80,9 @@ from packages.valory.skills.market_creation_manager_abci.models import (
 )
 from packages.valory.skills.market_creation_manager_abci.payloads import (
     DepositDaiPayload,
+    PostTxPayload,
     RemoveFundingPayload,
-    SyncMarketsPayload, PostTxPayload,
+    SyncMarketsPayload,
 )
 from packages.valory.skills.market_creation_manager_abci.rounds import (
     CollectRandomnessPayload,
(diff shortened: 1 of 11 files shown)


jmoreira-valory (approved):

packages/valory/skills/market_creation_manager_abci/behaviours.py
```diff
@@ -1388,6 +1390,110 @@ def async_act(self) -> Generator:
...
+            return PostTransactionRound.DONE_PAYLOAD
+
+        self.context.logger.info(f"Handling settled tx hash {settled_tx_hash}.")
+
+        if (
+            self.synchronized_data.tx_sender
+            != PrepareTransactionBehaviour.matching_round.round_id
+        ):
+            # we only handle market creation txs atm, any other tx, we don't need to take action
+            return PostTransactionRound.DONE_PAYLOAD
```
Adamantios, line 1429: `DONE_PAYLOAD` is actually not necessary as seen in the round's implementation. We could simply return `True` or `False` for done or error. The round would be a voting round, and by using the corresponding attributes we would not need to override the `end_block` method. Making use of these features of the framework would simplify both the behaviour and the round implementations.

Adamantios (approved):
Left a comment. Up to you if you would like to address it. Otherwise, feel free to merge.

packages/valory/skills/market_creation_manager_abci/behaviours.py
0xArdi: @Adamantios my idea was that this could have different payloads in the future depending on which round submitted the transaction.

PR #81, opened 2024-01-02
Market approval fix

jmoreira-valory:
Fixes market approval logic.

Now it examines all closing dates between `now` and `now + approve_market_event_days_offset` and ensures that there are `markets_to_approve_per_day` on each closing day in that interval.


packages/valory/skills/market_creation_manager_abci/behaviours.py
Adamantios, line 1: I recommend converting all these new logs to the debug level.

Adamantios (approved):

packages/valory/skills/market_creation_manager_abci/behaviours.py
jmoreira-valory: Thanks, will be considered for future improvement. The logs of this part are not very heavyweight now (only a few variables are displayed, and raw output of LLM and API calls have been suppressed in this part).

Issue #82, opened 2024-01-05
Agent gets stuck on `post_transaction_round` to update market address on market approval server.


jmoreira-valory:
Agent gets stuck on `post_transaction_round` to update market address on market approval server.

```
2024-01-05 13:29:21.924	
2024-01-05T12:29:21.779782933Z stdout F [2024-01-05 12:29:21,779] [INFO] [agent] 'post_transaction_round' round is done with event: Event.ERROR
2024-01-05 13:29:21.924	
2024-01-05T12:29:21.759143865Z stdout F [2024-01-05 12:29:21,759] [INFO] Created a new local deadline for the next `begin_block` request from the Tendermint node: 2024-01-05 12:30:21.759055
2024-01-05 13:29:21.924	
2024-01-05T12:29:21.759067162Z stdout F [2024-01-05 12:29:21,759] [INFO] [agent] no pending timeout, move time forward



2024-01-05 13:29:21.924	
2024-01-05T12:29:21.759045389Z stdout F [2024-01-05 12:29:21,758] [INFO] [agent] current AbciApp time: 2024-01-05 12:29:19.710402
2024-01-05 13:29:21.924	
2024-01-05T12:29:21.758791104Z stdout F [2024-01-05 12:29:21,758] [INFO] [agent] arrived block with timestamp: 2024-01-05 12:29:20.729118
2024-01-05 13:29:20.922	
2024-01-05T12:29:20.89855559Z stdout F ,message_id=-1,performative=error,target=1)
2024-01-05 13:29:20.922	
2024-01-05T12:29:20.898550206Z stdout F IndexError: tuple index out of range
2024-01-05 13:29:20.922	
2024-01-05T12:29:20.898544181Z stdout F 
2024-01-05 13:29:20.922	
2024-01-05T12:29:20.898540725Z stdout F             ~~~~^^^
2024-01-05 13:29:20.922	
2024-01-05T12:29:20.898537704Z stdout F     event = logs[0]
2024-01-05 13:29:20.922	
2024-01-05T12:29:20.898534146Z stdout F   File "/root/agent/vendor/valory/contracts/fpmm_deterministic_factory/contract.py", line 224, in parse_market_creation_event
2024-01-05 13:29:20.922	
2024-01-05T12:29:20.898528731Z stdout F 
2024-01-05 13:29:20.922	
[...]

PR #83, opened 2024-01-08
Fix code stuck on post transaction state

jmoreira-valory:
Fix code stuck on post transaction state

Addresses https://github.com/valory-xyz/market-creator/issues/82


packages/valory/skills/market_creation_manager_abci/rounds.py
```diff
@@ -771,7 +771,7 @@ class MarketCreationManagerAbciApp(AbciApp[Event]):
     transition_function: AbciAppTransitionFunction = {
         PostTransactionRound: {
             Event.DONE: CloseMarketsRound,
-            Event.ERROR: PostTransactionRound,
+            Event.ERROR: CloseMarketsRound,
```
0xArdi, line 774: Why would we end up in a situation where we cannot execute the logic in the PostTransaction cannot be executed properly?
Seems like the issue is elsewhere. Eitherway, it makes sense that we transition outside this round.

0xArdi (approved):

packages/valory/skills/market_creation_manager_abci/rounds.py
jmoreira-valory: This happened on #82, the service remained stuck there undefinitely (I had to restart it).

PR #84, opened 2024-01-31
Add lock check to workflow

angrybayblade:


jmoreira-valory (approved):

.github/workflows/main_workflow.yml
```diff
@@ -33,6 +35,8 @@ jobs:
         virtualenvs-in-project: false
         virtualenvs-path: ~/my-custom-path
         installer-parallel: true
+    - name: Poetry install requirements and check they can be locked
+      run: poetry install --all-extras --no-interaction -vvv
```
Adamantios, line 39: The install command reads the pyproject.toml file from the current project, resolves the dependencies, and installs them. If there is a poetry.lock file in the current directory, it will use the exact versions from there instead of resolving them.

Therefore, here we do not check if we can lock, but we check if we can install the lock deps.

angrybayblade: Fixed https://github.com/valory-xyz/market-creator/pull/84/commits/543ffc46c28a7ee2272241380ae0e123f85b33d5

Adamantios (approved):

PR #90, opened 2024-02-15
Integrate resolve market reasoning tool

jmoreira-valory:
Integrate resolve market reasoning tool


jmoreira-valory:
Implemented using mech_interact_abci on https://github.com/valory-xyz/market-creator/pull/91

PR #91, opened 2024-02-21
Resolve markets with mech requests

jmoreira-valory:
Merge main branch (bump)
Resolve markets with mech requests


packages/valory/skills/market_creation_manager_abci/behaviours.py
```diff
@@ -2443,14 +2285,26 @@ def _get_balance(self, account: str) -> Generator[None, None, Optional[int]]:
...
+            return GetPendingQuestionsRound.ERROR_PAYLOAD
+
+        filtered_questions = [
+            question
+            for question in unanswered_questions
+            if question["question"]["id"].lower()
+            not in self.shared_state.questions_requested_mech
+        ]
+        random.seed(self.last_synced_timestamp)
+        num_questions = min(len(filtered_questions), self.params.multisend_batch_size)
```
Adamantios, line 2301: @jmoreira-valory 

The `params` property does not specify the `multisend_batch_size`.

packages/valory/skills/market_creation_manager_abci/skill.yaml
```diff
@@ -178,6 +181,8 @@ models:
       max_attempts: 10
       max_healthcheck: 120
       multisend_address: '0x0000000000000000000000000000000000000000'
+      multisend_batch_size: 5
```
Adamantios, line 184: Instead, subclass from `MechParams`.

```suggestion
```

jmoreira-valory: Solved by adding a dummy variable on Params.

packages/valory/skills/market_creation_manager_abci/behaviours.py
```diff
@@ -2237,190 +2250,19 @@ def _get_unanswered_questions(
         response = yield from self.get_subgraph_result(
             query=OPEN_FPMM_QUERY.substitute(
                 creator=creator,
-                first=self.params.questions_to_close_batch_size,
+                first=1000,
```
Adamantios, line 2253: Why this change?

packages/valory/skills/market_creation_manager_abci/behaviours.py
```diff
@@ -2460,32 +2314,116 @@ def get_payload(self) -> Generator[None, None, str]:
...
 
-        # get the answers for those questions
-        question_to_answer = {}
+        # Prepare the Mech Requests for these questions
+        new_mech_requests = []
         for question in questions:
-            question_id = question["question"]["id"]
-            if question_id.lower() in self.shared_state.processed_question_ids:
+            question_id = question["question"]["id"].lower()
+            if question_id in self.shared_state.questions_requested_mech:
```
Adamantios, line 2338: Why are the `questions_requested_mech` not in the synced db?

packages/valory/skills/market_creation_manager_abci/behaviours.py
```diff
@@ -2460,32 +2314,116 @@ def get_payload(self) -> Generator[None, None, str]:
...
+            question_id = response.nonce
+            self.context.logger.info(
+                f"Received mech response: {response.nonce} {response.result}"
+            )
+
+            if question_id in self.shared_state.questions_responded:
+                continue
+
+            if question_id not in self.shared_state.questions_requested_mech:
+                continue
```
Adamantios, line 2420:
```suggestion
            if question_id in self.shared_state.questions_responded or question_id not in self.shared_state.questions_requested_mech:
                continue
```

packages/valory/skills/market_creation_manager_abci/behaviours.py
jmoreira-valory: Because 1000 is the maximum value returned by TheGraph (more than enough to collect all the expected unanswered questions), and `questions_to_close_batch_size` is expected to be a small number (in the range of 1-5). If these questions cannot be closed (e.g., mech returns error for them), there will be the problem that no more questions will be selected from the set of pending questions to close.

packages/valory/skills/market_creation_manager_abci/behaviours.py
jmoreira-valory: This was the similar approach designed previously. Since it's a single-agent service, it's easier to use the shared state in this case.

packages/valory/skills/market_creation_manager_abci/behaviours.py
Adamantios: Makes sense, can we add a constant though?

jmoreira-valory: I'm including it as part of the TheGraph query string, as it happens for the rest of the queries for this service.

Adamantios (approved):

PR #96, opened 2024-03-12
Add DepositDaiRound as first round

jmoreira-valory:
Add DepositDaiRound as first round. This will prevent the agent from getting stuck if there is not enough WxDAI when it's restarted.

### DIFF
--- a/scripts/list_finalizing_markets.ipynb
+++ b/scripts/list_finalizing_markets.ipynb
@@ -106,7 +106,7 @@
     "        timestamp = int(timestamp)\n",
     "        return datetime.utcfromtimestamp(timestamp).isoformat()\n",
     "    except Exception:  # pylint: disable=broad-except\n",
-    "        return \"\"\n",
+    "        return \"--\"\n",
     "\n",
     "\n",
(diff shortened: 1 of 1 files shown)


packages/valory/skills/market_creation_manager_abci/rounds.py
```diff
@@ -333,6 +336,9 @@ def end_block(self) -> Optional[Tuple[BaseSynchronizedData, Event]]:
             if self.most_voted_payload == self.REDEEM_BOND_DONE_PAYLOAD:
                 return self.synchronized_data, Event.REDEEM_BOND_DONE
 
+            if self.most_voted_payload == self.DEPOSIT_DAI_DONE_PAYLOAD:
+                return self.synchronized_data, Event.DEPOSIT_DAI_DONE
```
Adamantios, line 340: `DEPOSIT_DAI_DONE` missing from `PostTransactionRound`.

jmoreira-valory: You mean in transition function, right?
Fixed

Adamantios (approved):

PR #100, opened 2024-03-20
Update finalizing markets script

jmoreira-valory:
Fix aea-config.yaml


packages/valory/agents/market_maker/aea-config.yaml
```diff
@@ -61,8 +61,8 @@ logging_config:
     logfile:
       class: logging.FileHandler
       formatter: standard
-      filename: ${LOG_FILE:str:log.txt}
-      level: ${LOG_LEVEL:str:INFO}
+      filename: ${str:log.txt}
+      level: ${str:INFO}
```
jmoreira-valory, line 65: @angrybayblade Is this correct?

angrybayblade: Yes, but what was wrong with the old config?

jmoreira-valory: Is the format 
```
${LOG_FILE:str:log.txt}
```

acceptable in the aea-config.yaml file?  If so, I will revert the change. (I assumet it was only reserved to the service.yaml file)

PR #111, opened 2024-07-22
Upgrade reality.io bond from $0.1 to $10

evangriffiths:
Once market-creator uses the new OFV mech (https://github.com/valory-xyz/mech/pull/247), market resolution accuracy should go from 70% -> 90%. We hope to correct the remaining errors by incentivising Kleros community participation in resolutions by posting sufficiently large bonds (10 xDai)

Note: this will result in the deployed market-creator now losing bonds for incorrect resolutions. How this will be funded is being discussed separately 😃.

### DIFF
--- a/packages/valory/agents/market_maker/aea-config.yaml
+++ b/packages/valory/agents/market_maker/aea-config.yaml
@@ -207,7 +207,7 @@ models:
       use_termination: ${bool:false}
       use_slashing: ${bool:false}
       slash_cooldown_hours: ${int:3}
-      close_question_bond: ${int:1000000000000000}
+      close_question_bond: ${int:100000000000000000}
       questions_to_close_batch_size: ${int:1}
       slash_threshold_amount: ${int:10000000000000000}
       light_slash_unit_amount: ${int:5000000000000000}
--- a/packages/valory/services/market_maker/service.yaml
+++ b/packages/valory/services/market_maker/service.yaml
@@ -86,7 +86,7 @@ models:
       tendermint_check_sleep_delay: ${TENDERMINT_CHECK_SLEEP_DELAY:int:3}
       tendermint_com_url: ${TENDERMINT_COM_URL:str:http://localhost:8080}
       tendermint_max_retries: ${TENDERMINT_MAX_RETRIES:int:5}
-      close_question_bond: ${CLOSE_QUESTION_BOND:int:1000000000000000}
+      close_question_bond: ${CLOSE_QUESTION_BOND:int:100000000000000000}
       questions_to_close_batch_size: ${QUESTIONS_TO_CLOSE_BATCH_SIZE:int:1}
       tendermint_url: ${TENDERMINT_URL:str:http://localhost:26657}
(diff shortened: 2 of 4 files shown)


jmoreira-valory:
@evangriffiths There is no need to modify parameters in the YAML files. Environment variables can be set up accordingly when deploying the service. 

Please note the distinction between the answer bond and the answer bounty in Realitio.

(There is a refactor of the variable going on on https://github.com/valory-xyz/market-creator/pull/112)

evangriffiths:
Thanks for the review @jmoreira-valory

> Please note the distinction between the answer bond and the answer bounty in Realitio. (There is a refactor of the variable going on on https://github.com/valory-xyz/market-creator/pull/112)

Thanks for pointing that out. I'll update this PR if that is merged in first.

> There is no need to modify parameters in the YAML files.

I wasn't sure which yaml file does what, so I just changed them all 😅. Are you saying I should keep only the changes to the skill.yaml files, or that this PR isn't required at all, and the deployed market-creator can just be redeployed with new env vars?

While we're here, I'd appreciate it if you could explain why this is specified in 4 places 🙏.

jmoreira-valory:
@evangriffiths Yes, there is no need to change the YAML files, PR #112 is enough.

The reason why this constant is defined in 4 files is because the way the framework overrides dependencies:

Skill market_creation_manager is where the actual code is executed. The chain of dependencies is as follows:

1. `market_creation_manager_abci` defines a parameter.
2. `market_maker_abci` is a skill that deppends on the skill `market_creation_manager_abci` - to override a parameter from a dependency, the framework requires that the YAML file defines at this level the overridable parameter.
3. agent `market_maker` uses `market_maker_abci` - for the same reason, it needs to define here the overridable parameter.
4. similarly, for service `market_maker`.

If at any point 1-4 the parameter is missing, the "overrideability" will not be propagated down the chain.

Additionally, an overridable parameter on a service can define an .env variable (in upper case). Thus, this is how it will work:

1. user defines env variable, which is picked up by the framework when building the service deployment
2. the env variable is propagated to the dependencies until reaching the corresponding component.

For this reason, it's not necessary to change the values on the YAML files, as they can be changed "in real time" when creating the deployment to be executed.

Hopes this clarifies your question.

evangriffiths:
Okay, I will close in that case. Yes that makes sense, thanks for the explanation. But it's not possible to have 'required args' at the aea-config.yaml or abci/skill.yaml level? i.e. for the value to be defined just at the service level, like:

```
def market_creation_manager_abci(close_question_bond: int): pass

def market_maker_abci(close_question_bond: int):
    market_creation_manager_abci(close_question_bond=close_question_bond)

def market_maker_agent(close_question_bond: int):
    market_maker_abci(close_question_bond=close_question_bond)

def market_maker_service(close_question_bond: int = 100000000000000000):
    market_maker_agent(close_question_bond=close_question_bond)
```

jmoreira-valory:
No, that's not supported currently by the framework (although would be nice). It requires the parameter to be defined down the chain of dependencies.

PR #116, opened 2024-09-16
Feat/healthcheck

dvilelaf:
* Adds a healthcheck to the service


dvilelaf:
Hey balaji, I do not have the capacity to review all exercises from all participants, so just link this on your log and tag us whenever something fails only. Thanks!

jmoreira-valory (approved):

PR #119, opened 2024-10-04
Change default Reality's timeout to 3 days

kongzii:
Answers on Reality are challenged with $10, but sometimes people don't correct them in a day, especially on weekends. Motivation is to give people more time to correct the answers.


dagacha:
This won't be merged becasue the proposed change has a negative effect on trader agents because it will delay the time until they can reuse their capital.

I suggest to leave this PR open anyway in case someone is interested in this change they can reuse the code.

cyberosa:
Due to incompatibility reasons.

PR #120, opened 2024-10-08
Move to Kleros arbirtator with smaller fee (0.6 ETH) and possibility of appeals

kongzii:
See https://github.com/gnosis/prediction-market-agent-tooling/issues/444 for more details.

To sum it up, the current arbitrator cost is 10 ETH because it will spawn 511 jurors, and there is no possibility of appeals.

On gnosis-ai channel on Discord, an issue was raised that no one is going to pay such a huge fee for our low-liquidity markets and it makes people worried that they won't be able to afford the arbitration, if someone richer is going to challenge them. Which also happened yesterday:  **https://t.me/kleros/180186**

This PR changes the arbitrator to the one that costs 0.6 ETH, it spawns 31 jurors at the beginning, but appeal is possible.


jmoreira-valory:
Thanks for the detail, will propose to integrate it.

jmoreira-valory:
@kongzii I am integrating this in a working PR #121  which is currently being tested to be launched in production.

PR #122, opened 2024-10-28
Update pyproject

jmoreira-valory:
Update pyproject


cyberosa:
Not required anymore. If needed the updates will be incorporated in more recent open PRs

PR #124, opened 2024-11-26
Use Kleros' new arbitrator contract

kongzii:
In [this PR](https://github.com/valory-xyz/market-creator/pull/120) we switched to ` Kleros arbirtator with smaller fee (0.6 ETH) `, which is great, however Kleros found out they have some problems with that contract. 

They fixed it in our code[ in this PR](https://github.com/gnosis/prediction-market-agent-tooling/pull/550), so I'm just replicating it here 🙏


jmoreira-valory:
Thanks for pointing this.

This PR will be updated with the correct hashes before being merged. Meanwhile, we will redeploy today the market creator with the updated environment variable `ARBITRATOR_CONTRACT=0x5562Ac605764DC4039fb6aB56a74f7321396Cdf2` so that markets are oppened with the correct contract.

PR #140, opened 2025-09-24
fix: added check for market resolution date

LOCKhart07:


packages/valory/skills/market_creation_manager_abci/behaviours.py
```diff
@@ -735,16 +735,20 @@ def _sender_act(self) -> Generator:
...
                     for market in proposed_markets.values():
+                        if not self._is_resolution_date_in_question(market):
+                            self.context.logger.error(
+                                f"Resolution date wrong or not found in question {market=}"
+                            )
+                            continue
                         yield from self._propose_and_approve_market(market)
-
-                    approved_markets_count = len(proposed_markets)
+                        approved_markets_count += 1
```
Skanislav, line 751: Calling the function does not guarantee that the market was approved. It can fail because of http request

packages/valory/skills/market_creation_manager_abci/behaviours.py
```diff
@@ -761,6 +765,14 @@ def _sender_act(self) -> Generator:
 
         self.set_done()
 
+    def _is_resolution_date_in_question(self, market: Dict) -> bool:
+        formatted_resolution_date = datetime.fromtimestamp(
+            market["resolution_time"], tz=timezone.utc
+        ).strftime("%B %d, %Y")
+        if formatted_resolution_date in market["question"]:
+            return True
+        return False
```
Skanislav, line 774: Can you add a unit test for this one and add cases that were listed in the original report about this problem?

Skanislav (commented):
For the next time can you still keep the hashes in a separate PR? same reasoning as in trader

packages/valory/skills/market_creation_manager_abci/behaviours.py
LOCKhart07: https://github.com/valory-xyz/market-creator/pull/141/commits/ca9d0799dbdb4820040c39170c80f9b76245853e

PR #141, opened 2025-09-25
Resolution date tests

LOCKhart07:


packages/valory/skills/market_creation_manager_abci/handlers.py
```diff
@@ -47,6 +50,7 @@
 
 
 ABCIHandler = BaseABCIRoundHandler
+HttpHandler = BaseHttpHandler
```
LOCKhart07, line 53: Needed, else the BaseBehaviourTest raises error

Skanislav (approved):

PR #132, opened 2025-06-05
Refactor/deprecate wxdai contract

Skanislav:

PR #147, opened 2025-12-31
Deprecate WXDAI contract

jmoreira-valory:
Deprecate WXDAI contract. This PR obsoletes PR #132.

### DIFF
--- a/packages/valory/skills/market_creation_manager_abci/behaviours.py
+++ b/packages/valory/skills/market_creation_manager_abci/behaviours.py
@@ -1035,8 +1035,6 @@ class RedeemBondBehaviour(MarketCreationManagerBaseBehaviour):
                 f"RealitioContract.build_withdraw_tx unsuccessful! : {response}"
             )
             return None
-
-        # data_hex = "0x" + response.state.body["data"].hex()
         return {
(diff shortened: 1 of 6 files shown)


packages/valory/skills/market_creation_manager_abci/behaviours.py
```diff
@@ -1399,8 +1397,8 @@ def _get_withdraw_tx(self, amount: int) -> Generator[None, None, Optional[Dict]]
         response = yield from self.get_contract_api_response(
             performative=ContractApiMessage.Performative.GET_STATE,
             contract_address=self.params.conditional_tokens_contract,
-            contract_id=str(WxDAIContract.contract_id),
-            contract_callable=get_callable_name(WxDAIContract.build_withdraw_tx),
+            contract_id=str(ERC20.contract_id),
+            contract_callable=get_callable_name(ERC20.build_withdraw_tx),
```
jmoreira-valory: The output format of `_get_withdraw_tx` (line 1409) does not need to be modified, even with the difference in interface between former WxDAI contract and ERC20 contract. This is because, downstream, the multisend contract correctly works for "0x" strings and bytes.

Adamantios (approved):

PR #188, opened 2026-04-08
Remove notebooks from repository

jmoreira-valory:
Remove notebooks from repository


LOCKhart07 (commented):
## Review

**Blocking**

1. **`_calc_prices` imported as private function** (`market_creator_activity.ipynb`)
   The notebook does `from subgraph_omen import _calc_prices`. This is a private function already called internally by `markets_to_dataframe`. If it gets renamed/refactored the notebook silently breaks. Consider making it public (`calc_prices`) if it's part of the intended API.

2. **QS and Pearl share the same icon** (`market_creator_activity.ipynb`)
   Both `qs` and `pearl` configs set `icon: "🏠"` (house). The comment above `MARKET_CREATORS` says each creator/resolver must declare a **unique** icon so the responder chain can visually distinguish them — with identical icons there's no way to tell QS from Pearl in the chain.

3. **Empty markdown cell** (`market_creator_activity.ipynb`)
   Cell id `a577f571` has `"source": []` — looks like a leftover, should be removed.

**Non-blocking**

4. **CPMM fallback uses integer division** (`subgraph_omen.py`, `_calc_prices`)
   `inv = [product // a for a in amounts]` — floor division. For large token amounts the rounding error is negligible, but `product / a` would be more precise. Unlikely to matter in practice.

🤖 Generated with [Claude Code](https://claude.ai/code)

LOCKhart07 (approved):

DhairyaPatel7 (approved):

LOCKhart07 (approved):

PR #189, opened 2026-04-09
Remove answering flow from market-creator

jmoreira-valory:
## Summary

Removes the question-responding flow from `market-creator` — the responsibility now lives in the separate `market-resolver` service. Market cleanup flows (`RemoveFundingRound`, `RedeemWinningsRound`) are untouched.

### What is removed

**Rounds / behaviours / states** (`market_creation_manager_abci`):
- `GetPendingQuestionsRound` / `Behaviour` / `state`
- `AnswerQuestionsRound` / `Behaviour` / `state`
- `RedeemBondRound` / `Behaviour` / `state`
- `Event.MECH_REQUEST_DONE`, `Event.REDEEM_BOND_DONE`, `Event.ANSWER_QUESTION_DONE`
- `FinishedWithRedeemBondRound`, `FinishedWithGetPendingQuestionsRound`, `FinishedWithAnswerQuestionsRound`, `FinishedWithMechRequestRound`
- `mech_requests`, `mech_responses` from `SynchronizedData`
- `GetPendingQuestionsPayload`
- `do_llm_request` helper on the base behaviour
- `LlmHandler` / `LlmDialogues`

**FSM rewiring**:
- `DepositDaiRound`: `NO_MAJORITY` / `NONE` / `ROUND_TIMEOUT` → `CollectRandomnessRound`
- `SelectKeeperRound.DONE` → `CollectProposedMarketsRound` (was `RedeemBondRound`)
- `PostTransactionRound`: dropped `MECH_REQUEST_DONE_PAYLOAD`, `REDEEM_BOND_DONE_PAYLOAD`, `ANSWER_QUESTION_DONE_PAYLOAD` branches

**Composed `market_maker_abci`**:
- Removed `MechInteractAbciApp` from the chain and all 11 related `abci_app_transition_mapping` entries
- Removed `MechInteractParams` from `Params` MRO and the `MechInteractEvent.ROUND_TIMEOUT` entry in `event_to_timeout`
- Removed `MechInteractRoundBehaviour` from `behaviours.py`
- Removed `Llm*` handlers/dialogues

**Config & manifests**:
- `aea-config.yaml` / `service.yaml`: removed all mech contracts, the `llm` protocol, the `mech_interact_abci` skill, the `valory/openai` connection + its config block, `mech_response` model, and every `mech_*` param plus `use_acn_for_delivers`, `agent_registry_address`, `realitio_answer_question_bond`, `questions_to_close_batch_size`, `answer_retry_intervals`, `irrelevant_tools`, `ignored_mechs`, `penalize_mech_time_window`, `mech_wrapped_native_token_address`, `deliveries_lookback_days`
- `skill.yaml` (both skills): same parameter pruning, removed `llm` protocol, removed `mech_interact_abci` dep, removed `mech_response` / `mech_tools` / `mechs_subgraph` models
- `packages/packages.json`: removed the corresponding `third_party` entries and pruned the orphaned local synced directories

### What stays

- **Question generation is unaffected.** `approve_markets.py` still calls `propose_questions.run()` in-process using `self.params.openai_api_key` — this never went through the `llm` protocol or the `valory/openai` AEA connection.
- `RedeemWinningsRound` and `RemoveFundingRound` (market cleanup) are kept along with `redeem_winnings_batch_size`.
- All still-referenced contracts are preserved: `realitio`, `realitio_proxy`, `conditional_tokens`, `collateral_tokens` (wxDAI), `fpmm`, `fpmm_deterministic_factory`, `gnosis_safe`, `multisend`.
- `realitio_answer_question_bounty` param is kept (still consumed by `prepare_transaction.py`).

### Test updates

Three test files were recreated mech-free after the original commit deleted them wholesale along with the responding-flow tests. They cover code paths that are still live (`PostTransactionBehaviour` / `PostTransactionRound` branches for `DepositDai` / `RemoveFunding` / `RedeemWinnings` / `funds_forwarder`, and `SharedState.setup()` in the composed app):

- `market_creation_manager_abci/tests/behaviours/test_post_transaction.py`
- `market_creation_manager_abci/tests/states/test_post_transaction.py`
- `market_maker_abci/tests/test_models.py`

707 unit tests passing locally at 100% coverage.

## Test plan

- [x] `tox -e isort` / `tox -e black`
- [x] `tox -e mypy` / `tox -e flake8` / `tox -e darglint` (pre-existing failures in `scripts/audit_trader_redemptions.py` are unrelated)
- [x] `tox -e check-abciapp-specs` / `check-handlers` / `check-packages`
- [x] `autonomy packages lock`
- [x] `tox -e unit-tests` — 707 passing at 100% coverage
- [x] Boot the agent against a fresh safe and confirm the market-creation → FPMM-deployment cycle completes without hitting any removed round


Adamantios (approved):

packages/valory/skills/market_maker_abci/composition.py
```diff
@@ -58,22 +53,9 @@
...
-    MechFinalStates.FinishedMechLegacyDetectedRound: MechRequestStates.MechRequestRound,
-    MechFinalStates.FinishedMechInformationRound: MechRequestStates.MechRequestRound,
-    MechFinalStates.FailedMechInformationRound: MechVersionStates.MechVersionDetectionRound,
-    MechFinalStates.FinishedMechRequestRound: TransactionSettlementAbci.RandomnessTransactionSubmissionRound,
-    MechFinalStates.FinishedMechPurchaseSubscriptionRound: TransactionSettlementAbci.RandomnessTransactionSubmissionRound,
-    MechFinalStates.FinishedMechResponseRound: MarketCreationManagerAbci.AnswerQuestionsRound,
-    MechFinalStates.FinishedMechRequestSkipRound: MarketCreationManagerAbci.CollectRandomnessRound,
-    MechFinalStates.FinishedMechResponseTimeoutRound: MarketCreationManagerAbci.CollectRandomnessRound,
-    MarketCreationManagerAbci.FinishedWithMechRequestRound: MechResponseStates.MechResponseRound,
     MarketCreationManagerAbci.FinishedWithRedeemWinningsRound: TransactionSettlementAbci.RandomnessTransactionSubmissionRound,
```
LOCKhart07, line 58: Heads up: `tests/test_composition.py` (182 lines) was deleted in this PR with no replacement. It was the only place asserting the contents of `abci_app_transition_mapping` — both the expected (source, target) pairs and the total mapping count. Combined with the deletion of `market_creation_manager_abci/tests/test_rounds.py` (572 lines, which asserted `transition_function`/`initial_states`/`final_states`/`db_pre_conditions`/etc.), there's now no semantic regression test for the FSM rewiring that this PR is doing.

This is invisible to `--cov-fail-under=100` because these are declarative dict literals — line coverage is trivially 100% from the import. And `tox -e check-abciapp-specs` only validates yaml↔rounds.py *consistency*, not whether the wiring matches intent. After this PR, a typo like swapping a target round in either `transition_function` or `abci_app_transition_mapping` would pass CI.

Not blocking, but worth restoring at least the parametrized expected-pairs check from `test_composition.py` and the equivalent from `test_rounds.py` for the kept rounds, given the FSM rewiring is the primary change here.

packages/valory/skills/market_creation_manager_abci/rounds.py
```diff
@@ -91,173 +78,124 @@ class MarketCreationManagerAbciApp(AbciApp[Event]):
...
-        },
         CollectRandomnessRound: {
             Event.DONE: SelectKeeperRound,
             Event.NO_MAJORITY: CollectRandomnessRound,
             Event.NONE: CollectRandomnessRound,
             Event.ROUND_TIMEOUT: CollectRandomnessRound,
         },
         SelectKeeperRound: {
-            Event.DONE: RedeemBondRound,
+            Event.DONE: CollectProposedMarketsRound,
```
LOCKhart07, line 194: Same coverage gap as flagged on `composition.py`: `tests/test_rounds.py` (572 lines) was deleted with no replacement, so the rewired transitions here (this `SelectKeeperRound.DONE` → `CollectProposedMarketsRound`, the `DepositDaiRound` NONE/NO_MAJORITY/ROUND_TIMEOUT → `CollectRandomnessRound` rewiring above, and the three branches removed from `PostTransactionRound`) are no longer asserted by any test.

LOCKhart07 (approved):
Approving — the FSM rewiring itself looks correct (composition mapping accounts for all 5 remaining `MarketCreationManagerAbciApp` final states, `rounds.py`/`fsm_specification.yaml`/`states/base.py` are internally consistent, and no leftover references to removed params/connections anywhere except `propose_questions.py` which is excluded from coverage). Two non-blocking comments inline.

Minor nit on the description: it says the three test files were "recreated mech-free after the original commit deleted them wholesale", but the diff shows they were modified, not recreated — `tests/behaviours/test_post_transaction.py` is −37/+0, `tests/states/test_post_transaction.py` is −14/+0, and `market_maker_abci/tests/test_models.py` is −20/+14. Just a description nit, not a code issue.

packages/valory/skills/market_creation_manager_abci/rounds.py
jmoreira-valory: After discussing, we decided NOT to recreate test_composition.py / test_rounds.py. The prior versions were mirror-assertion tests: the "expected" data was copy-pasted from the source file (abci_app_transition_mapping in composition.py, transition_function / initial_states / final_states in rounds.py) and the tests asserted that the source equals its own copy.

That form of test doesn't catch any bug that fsm_specification.yaml + tox -e check-abciapp-specs doesn't already catch — and the latter is auto-regenerated from rounds.py via autonomy analyse fsm-specs --update, so there's a single source of truth instead of two hand-maintained mirrors. A transition change that breaks CI will fail at check-abciapp-specs regardless.

The semantic coverage that DOES matter (full FSM execution) lives in the behaviour tests, which exercise rounds end-to-end. We'll keep that surface covered going forward.

If a specific higher-level invariant would be worth asserting in a unit test (e.g., "every final state is reachable", "no orphan events in the enum"), happy to add those in a follow-up - they'd be real invariants rather than source copies.

PR #195, opened 2026-04-16
feat: improve question generation pipeline — reduce skew and divergence

jmoreira-valory:
Important: this PR has merged #197 on it - reviewers should only address changes in the skill `market_creation_manager_abci ` 

## Summary

Improves `propose_questions.py` to reduce two problems identified in the Jury vs Crowd v2 analysis:

1. **Yes/No resolution skew** — Omen resolves 16% Yes recently vs ~30% baseline
2. **Jury–crowd divergence** — 24.5% disagreement, 42% from fixable question-framing traps

## Changes to `propose_questions.py`

- **Two-step generation** — new `EXTRACT_STATE_PROMPT` identifies measurable states from articles before question framing
- **Chain-of-thought self-review** — forces explicit deadline arithmetic before accept/reject
- **Programmatic date validation** — Python-level check catches past/out-of-range dates (100% recall)
- **Generate 3 candidates, keep best** — self-review as selector, not just gate
- **Explicit window parameter** — prompt receives `WINDOW: N days`
- **Moderation pre-filter** — per-article instead of per-prompt; fixes 72% blocking rate
- **No fallback to rejected questions** — returns error instead of shipping failed questions

## Benchmark harness

`benchmark_propose_questions.py` — parallel execution (ThreadPoolExecutor + tqdm), subgraph-seeded dedup, JSONL output.

## Results so far

| Metric | Baseline | Current |
|---|---|---|
| Non-announcement questions | 0% | **62%** |
| Measurement framings | 0% | **31%** |
| Continuation framings | 0% | **22%** |
| Trap risk | 46% | **15%** |
| Expected Yes resolution | 0% | **28%** |
| Moderation throughput | 28% | **100%** |

Work in progress — further iterations on this branch.


packages/valory/skills/market_creation_manager_abci/propose_questions.py
```diff
@@ -167,6 +333,41 @@ class LLMStorySelectionSchema(BaseModel):
...
+                try:
+                    d = datetime.strptime(match.replace(",", ""), fmt.replace(",", ""))
+                    d = d.replace(tzinfo=timezone.utc)
+                    if d < now - __import__("datetime").timedelta(days=1):
+                        return f"Date '{match}' is in the past"
+                    if d > deadline + __import__("datetime").timedelta(days=365):
+                        return f"Date '{match}' is too far beyond the deadline"
+                    break
+                except ValueError:
+                    continue
```
Adamantios: This is hard to read. Also, why the unconventional `__import__("datetime")`?

packages/valory/skills/market_creation_manager_abci/propose_questions.py
```diff
@@ -167,6 +333,41 @@ class LLMStorySelectionSchema(BaseModel):
     reasoning: str
 
 
+import re
+
```
Adamantios: Why here?

jmoreira-valory: fixed

packages/valory/skills/market_creation_manager_abci/propose_questions.py
jmoreira-valory: fixed

packages/valory/skills/market_creation_manager_abci/propose_questions.py
```diff
@@ -575,8 +910,112 @@ def run(**kwargs: Any) -> Tuple[Optional[str], Optional[Dict[str, Any]], Any, An
                 )
             response = json.loads(response.choices[0].message.content)
 
-        # Generate output
+        # Self-review pass: audit proposed questions against the four trap checks.
+        # Questions that fail any check are rejected before they reach the output.
         questions = response["questions"][:num_questions]
```
LOCKhart07, line 915: This truncates to `num_questions` **before** the self-review and date-validation passes. The prompt asks the LLM for `n_candidates = max(num_questions * 5, 5)` (line 857) specifically so self-review can act as a selector; with this slice self-review only ever sees `num_questions` candidates (typically 1), which contradicts the PR description's "generate 3 candidates, keep best" and makes the `All N proposed questions were rejected` error far more likely (a single bad candidate now takes the whole call down). Fix: drop the slice (`questions = response["questions"]`); `date_validated[:num_questions]` at line 1009 already bounds the final output.

packages/valory/services/market_maker/service.yaml
```diff
@@ -89,7 +71,7 @@ models:
       tendermint_p2p_url: ${TENDERMINT_P2P_URL:str:localhost:26656}
       termination_sleep: ${TERMINATION_SLEEP:int:900}
       termination_from_block: ${TERMINATION_FROM_BLOCK:int:0}
-      xdai_threshold: ${XDAI_THRESHOLD:int:3000000000000000000}
+      xdai_threshold: ${XDAI_THRESHOLD:int:100000000000000000}
```
LOCKhart07, line 74: `XDAI_THRESHOLD` default silently changed from `3000000000000000000` (3 xDAI) to `100000000000000000` (0.1 xDAI) — a 30× reduction not mentioned in the PR description. With the deposit logic in `behaviours/deposit_dai.py:90-96` (`balance_to_deposit = balance - xdai_threshold`), this leaves much less xDAI in the Safe for gas on multisend/redemption txs. Also, both `skill.yaml` files still carry `3e18`, so only the service override drifts — skill-level defaults are inconsistent. Intentional?

packages/valory/skills/market_creation_manager_abci/propose_questions.py
```diff
@@ -575,8 +910,112 @@ def run(**kwargs: Any) -> Tuple[Optional[str], Optional[Dict[str, Any]], Any, An
...
+            )
+            review_messages = [
+                {
+                    "role": "system",
+                    "content": "You are a prediction-market question auditor.",
+                },
+                {"role": "user", "content": review_prompt},
+            ]
+            review_response = client.chat.completions.create(
+                model=model,
```
LOCKhart07, line 933: `model` here is implicitly inherited from the question-generation block above (where it was set to `TOOL_TO_ENGINE[tool]` = gpt-4.1). Works, but it relies on leak-of-scope; if the intent is to use the stronger model for review, bind it explicitly (`model = TOOL_TO_ENGINE[tool]`) so the choice isn't accidental. Otherwise this silently drifts if someone later rebinds `model` inside the generation block.

packages/valory/skills/market_creation_manager_abci/propose_questions.py
```diff
@@ -575,8 +910,112 @@ def run(**kwargs: Any) -> Tuple[Optional[str], Optional[Dict[str, Any]], Any, An
...
+        # Self-review pass: audit proposed questions against the four trap checks.
+        # Questions that fail any check are rejected before they reach the output.
         questions = response["questions"][:num_questions]
+
+        with OpenAIClientManager(kwargs["api_keys"]["openai"]):
+            assert client is not None
+            review_prompt = SELF_REVIEW_PROMPT.format(
+                questions=json.dumps(questions, indent=2),
+                latest_questions=latest_questions_string,
+                article=f"{scrape_result['text'][:4000]}",
```
LOCKhart07, line 922: Article text is truncated to `[:4000]` here but `[:6000]` at line 802 for the generation prompt. The reviewer is scoring a shorter article than the generator saw. Align the caps (or pull out a named constant).

packages/valory/skills/market_creation_manager_abci/propose_questions.py
```diff
@@ -575,8 +910,112 @@ def run(**kwargs: Any) -> Tuple[Optional[str], Optional[Dict[str, Any]], Any, An
...
+            if date_issue:
+                rejected_questions.append(
+                    {"question": q, "reason": f"DATE CHECK: {date_issue}"}
+                )
+            else:
+                date_validated.append(q)
+
+        print(
+            f"Self-review: {len(accepted_questions)} accepted, "
+            f"{len(rejected_questions)} rejected out of {len(questions)} proposed"
```
LOCKhart07, line 1001: `len(questions)` here reports the post-truncation count, so this always prints `out of num_questions proposed` rather than the `n_candidates` actually requested from the LLM. Misleading when diagnosing low-yield cycles — use the original `n_candidates` instead.

packages/valory/services/market_maker/service.yaml
jmoreira-valory: This is irrelevant.

packages/valory/skills/market_creation_manager_abci/propose_questions.py
jmoreira-valory: fixed

packages/valory/skills/market_creation_manager_abci/propose_questions.py
jmoreira-valory: fixed

packages/valory/skills/market_creation_manager_abci/propose_questions.py
jmoreira-valory: fixed

packages/valory/skills/market_creation_manager_abci/propose_questions.py
```diff
@@ -571,12 +1080,147 @@ def run(**kwargs: Any) -> Tuple[Optional[str], Optional[Dict[str, Any]], Any, An
...
+            questions = nondup_validated[:num_questions]
+        else:
+            err_payload = {
+                "error": (
+                    f"All {n_candidates} proposed questions were rejected by "
+                    "self-review, date validation, or programmatic dedup."
+                ),
+                "tool": tool,
+            }
+            return json.dumps(err_payload), None, None, counter_callback
```
LOCKhart07, line 1261: This new error path correctly uses `json.dumps(...)`, but the seven older error returns in this file (at L790, L799, L816, L835, L878, L913, L1052, L1248) still use `f'{{"error": "...", "tool": {tool}}}'` — `{tool}` is interpolated unquoted, so the output becomes e.g. `{"error": "...", "tool": propose-question}` — not parseable. `approve_markets.py:144` does `json.loads(mech_tool_output)` with no try/except, so any of those paths will crash `_sender_act` on the keeper. Apply the same `json.dumps` fix across the board.

packages/valory/skills/market_creation_manager_abci/behaviours/approve_markets.py
```diff
@@ -144,16 +144,17 @@ def _sender_act(self) -> Generator:
                 mech_tool_output_json = json.loads(mech_tool_output)
                 # END MECH INTERACT EMULATION
 
-                self.context.logger.info(f"{mech_tool_output_json['reasoning']=}")
-
-                proposed_markets = mech_tool_output_json["questions"]  # type: ignore
-
                 approved_markets_count = 0
```
LOCKhart07: Dead re-assignment — `approved_markets_count = 0` is already set on L112 before the `if opening_ts:` block. Drop this line.

LOCKhart07 (dismissed):

DhairyaPatel7 (dismissed):

packages/valory/skills/market_creation_manager_abci/propose_questions.py
jmoreira-valory: fixed

packages/valory/skills/market_creation_manager_abci/behaviours/approve_markets.py
jmoreira-valory: fixed

DhairyaPatel7 (approved):

Adamantios (approved):

PR #196, opened 2026-04-16
feat: add ct_redeem_tokens_min_payout and ignored positions optimization

jmoreira-valory:
## Summary

Adds `ct_redeem_tokens_min_payout` parameter (default: 1 wei) and an in-memory `ignored_ct_positions` set to the `omen_ct_redeem_tokens_abci` skill. With `min_payout > 0`, losing positions are skipped and cached in `ignored_ct_positions` on SharedState, eliminating ~87 redundant subgraph queries per cycle for the QS safe's 7,855 dead positions. Setting `CT_REDEEM_TOKENS_MIN_PAYOUT=0` in deployment triggers a one-time burn of all losing tokens (~1.7 xDAI gas), permanently cleaning the CT subgraph.

## Test plan

- [x] 111 unit tests pass
- [x] Linters pass (isort, black, pylint, mypy, flake8, darglint, bandit)
- [x] Package hashes locked
- [x] Deploy with default (min_payout=1) and verify losers are skipped + ignored set grows
- [x] Deploy with min_payout=0 and verify losers are burned


packages/valory/skills/omen_ct_redeem_tokens_abci/behaviours/redeem_tokens.py
```diff
@@ -120,18 +120,29 @@ def async_act(self) -> Generator:
...
         """Run the full CT redemption pipeline as a linear recipe."""
-        # Step 1: find what the safe holds
+        # Step 1: find what the safe holds (excluding ignored positions)
         held_positions = yield from self._get_held_positions()
+        ignored = self.shared_state.ignored_ct_positions
+        if ignored:
+            held_positions = {
+                cid: idx_sets
+                for cid, idx_sets in held_positions.items()
+                if cid not in ignored
```
Adamantios: The ignored are converted to lower before being added, so it's worth adding a `lower` here as well for safety:

```suggestion
                if cid.lower() not in ignored
```

packages/valory/agents/market_maker/aea-config.yaml
```diff
@@ -222,6 +222,7 @@ models:
       liquidity_removal_lead_time: ${int:86400}
       fpmm_remove_liquidity_batch_size: ${int:1}
       ct_redeem_tokens_batch_size: ${int:5}
+      ct_redeem_tokens_min_payout: ${int:1}
```
Adamantios, line 225: Isn't the `min_payout` part a bit misleading? In essence, 1 vs 10 would have the same result, no?

jmoreira-valory: Yes, let me explain:

CT positions appears as "redeemable" in subgraph/onchain EVEN if they have 0 value. - It makes no sense redeem 0-value CT tokens, it's just a waste of gas, BUT on the other hand it will "clear" the status for these and won't appear in the queries as pending CT.

For this reason, is up to the user define this parameter to 0 and "clear" 0-value CT positions. Alternatively, set it to a reasonable "DUST THRESHOLD" like the trader does.

For simplicity, we simply exclude here by default 0-value CT positions, but it's configurable by the final user.

packages/valory/skills/omen_ct_redeem_tokens_abci/behaviours/redeem_tokens.py
jmoreira-valory: fixed

packages/valory/skills/omen_ct_redeem_tokens_abci/behaviours/redeem_tokens.py
```diff
@@ -155,22 +166,45 @@ def _prepare_multisend(self) -> Generator[None, None, Optional[str]]:
...
+
+        :param held_positions: mapping of condition_id to held index sets.
+        :yield: None
+        :return: list of redeemable market dicts.
+        """
         positions_to_redeem = yield from self._get_markets_for_conditions(
             list(held_positions.keys())
         )
 
+        burn_losers = self.params.ct_redeem_tokens_min_payout == 0
```
LOCKhart07, line 187: `ct_redeem_tokens_min_payout` is effectively a boolean toggle, not a threshold. `burn_losers = self.params.ct_redeem_tokens_min_payout == 0` means values of 1, 10, or 1_000_000 all behave identically, and `_has_winning_position` still compares `float(payout) > 0` rather than against `min_payout`. The param name and your reply in the earlier thread ("set to a reasonable DUST THRESHOLD like the trader does") both imply threshold semantics that aren't actually implemented, which will confuse future operators who set it to e.g. 10. Two options:

- Rename to something like `burn_losing_ct_positions: bool` and switch the type in skill.yaml / aea-config.yaml / service.yaml.
- Actually thread the threshold through, e.g. `float(payout) >= min_payout` inside `_has_winning_position`.

Related nits worth addressing while you're here:
- The PR description says min_payout=0 "triggers a one-time burn of **all** losing tokens", but `_get_markets_for_conditions` filters out markets whose payouts are all zero, so those positions never reach the burn path. Probably not reachable in practice for Omen, but the claim isn't literal.
- Once a condition lands in `ignored_ct_positions` it's never removed. If a Realitio answer is overturned via arbitration and the winner flips, the now-winning position stays permanently skipped until the agent restarts. Worth a docstring caveat (or clearing the ignored set on some event).

packages/valory/skills/omen_ct_redeem_tokens_abci/tests/behaviours/test_redeem_tokens.py
```diff
@@ -422,6 +474,33 @@ def test_prepare_multisend_no_held(self) -> None:
...
+    def test_prepare_multisend_partial_ignored(self) -> None:
+        """Test _prepare_multisend filters out ignored positions."""
+        held = {"0xcond1": {1}, "0xcond2": {2}}
+        self.behaviour.context.state.ignored_ct_positions = {"0xcond1"}
+        markets = [{"address": "0xM2", "condition_id": "0xcond2"}]
+        txs = [{"to": "0xCT", "data": "0xredeem", "value": 0}]
+        with (
+            patch.object(self.behaviour, "_get_held_positions", new=make_gen(held)),
+            patch.object(
+                self.behaviour, "_get_redeemable_markets", new=make_gen(markets)
```
LOCKhart07: This test doesn't actually verify the new filter in `_prepare_multisend`. `_get_redeemable_markets` is mocked with `make_gen(markets)` that ignores its arguments, so the assertion `result == "0xhash"` would pass even if the `ignored` filter were deleted. Per `~/.claude/rules/testing.md`, this is testing the mock, not the code.

Suggest capturing the call args (e.g. wrap `_get_redeemable_markets` with a `MagicMock` side_effect that records its input) and asserting it receives `{"0xcond2": {2}}`, not the full `held`. `test_prepare_multisend_all_ignored` is falsifiable and fine as-is.

packages/valory/skills/omen_ct_redeem_tokens_abci/behaviours/redeem_tokens.py
```diff
@@ -120,18 +120,29 @@ def async_act(self) -> Generator:
...
         """Run the full CT redemption pipeline as a linear recipe."""
-        # Step 1: find what the safe holds
+        # Step 1: find what the safe holds (excluding ignored positions)
         held_positions = yield from self._get_held_positions()
+        ignored = self.shared_state.ignored_ct_positions
+        if ignored:
+            held_positions = {
+                cid: idx_sets
+                for cid, idx_sets in held_positions.items()
+                if cid.lower() not in ignored
```
LOCKhart07, line 130: `cid.lower()` here is redundant — `_get_held_positions` already lowercases all keys before inserting (see `cid_lower = cid.lower()` in that method). Keeping the `.lower()` call reads as if the key case is unknown, which muddies the invariant. Either drop it or (if you want defense-in-depth) add a comment noting it's belt-and-braces.

packages/valory/skills/omen_ct_redeem_tokens_abci/behaviours/redeem_tokens.py
jmoreira-valory: I disagree with these comments. ct_redeem_tokens_min_payout is the minimum amount the user wants to recover (eventually being set at 0).

packages/valory/skills/omen_ct_redeem_tokens_abci/behaviours/redeem_tokens.py
jmoreira-valory: Not changing it. See comment https://github.com/valory-xyz/market-creator/pull/196#discussion_r3099256816

packages/valory/skills/omen_ct_redeem_tokens_abci/tests/behaviours/test_redeem_tokens.py
jmoreira-valory: addressed

LOCKhart07 (approved):

PR #201, opened 2026-04-30
chore: Wave 2a + 2b — tomte canonical migration

DhairyaPatel7:
## Summary

Centralises tooling state through **tomte SHA `0819ee` (post-#46 merge)** and consumes its auto-derived defaults. Mirrors the Wave 2a/2b shape landed in [trader#925](https://github.com/valory-xyz/trader/pull/925), [mech#437](https://github.com/valory-xyz/mech/pull/437), and the sibling agent repos.

Includes the wave-2b dep bump on top:
- **OA 0.21.18 → 0.21.20**, **OAEA 2.2.1 → 2.2.3** (helpers, test-autonomy, cli-ipfs, ledger-cosmos, ledger-ethereum)
- **upstream_pins**: `mech-interact@v0.30.0`, `trader@v0.35.5`, `genai@v7.0.0`
- 38 third-party hashes refreshed against the bumped upstream tags
- `release.yml`: bump 5 OA refs (incl. 2× breaking `-e open-autonomy==0.21.18` `EXTRA_DEPENDENCIES`)

## Wave 2a — tomte tox runtime migration

Per-repo identity moves to `[tool.tomte]` in `pyproject.toml`; the long hand-written `tox.ini` shrinks to a `[tomte-extensions]` overlay consumed by `tomte tox` at runtime.

**Key overrides:**
- `upstream_pins` — third-party skills/contracts on disk synced from these upstream repos (open-autonomy/open-aea auto-derived from pyproject).
- `gitleaks_extra_regexes` — IPFS CIDv1 hashes plus content-based allowlists for ABCI `_PAYLOAD` constants, `participant_to_*` synced-data attrs, and `*_(contract|address|...)` config keys (refined per review — was previously a blanket `packages/valory/skills/.*` path allowlist).
- `gitleaks_extra_paths` — `market_approval_server/` + `.claude/` only.

**Lint debt fixed (zero `ignore_errors` for first-party):**
- mypy: 3 errors → 0 (added gql/eth_utils to `ignore_missing_imports`)
- pylint: 10.00/10
- 2 files black-reformatted, 1 isort-fixed

## Fleet-wide retirements (no longer in CI)

These were dropped in lockstep with kv-store, funds-manager, genai, mech, mech-agents-fun, mech-client, mech-interact, mech-predict, mech-server, trader, optimus, meme-ooorr:

[...]

PR #202, opened 2026-05-09
fix: trader v0.37.0 bump, omen workaround revert, graphql-core pin, deps/ports

jmoreira-valory:
## Summary

- **Bump `trader` upstream pin v0.35.5 → v0.37.0**
  - Carries the `simulate_claim_winnings` data-slice fix (valory-xyz/trader#950 + the blocking review on it)
  - Carries the bytes-by-construction migration for 4 dev contracts in trader (`realitio`, `realitio_proxy`, `conditional_tokens`, `relayer`)
- **Drop the defensive multisend workarounds** in the omen skills now that contracts emit bytes natively:
  - 3 nested `_calldata_to_bytes` helpers + 6 `.hex() if isinstance(...)` guards across `omen_realitio_withdraw_bonds_abci`, `omen_ct_redeem_tokens_abci`, `omen_fpmm_remove_liquidity_abci`
- **Pin `graphql-core>=3.2,<3.3`** (in `pyproject.toml`, `aea-config.yaml`, and `market_creation_manager_abci/skill.yaml`) to avoid an alpha (`3.3.0a12`) being resolved transitively, which froze AST nodes and crashed `gql` query building with `cannot assign to field 'selections'` during `propose-question` Mech tool execution. Also tightened `gql<3.6` to avoid pre-release pickups.
- **Add `event_filtering_batch_size` skill param** (default 1000) to chunk `eth_getLogs` in Realitio claim-params discovery — Gnosis RPC enforces log-range limits (rationale in `models.py:62-65`).
- **Set `http_server` port default `8716` → `8000`** to match Propel's k8s ingress (was a foot-gun requiring manual `HTTP_SERVER_PORT=8000` env var, otherwise healthcheck returned 502).
- **Tighten `test_get_claim_params_success`** to assert `chunk_size` kwarg is actually forwarded to the contract API call (post-review hardening).
- **Restore `hypothesis==6.21.6` to agent runtime deps** to recover the pre-#201 Propel deployment state — see section below.

## Why `hypothesis==6.21.6` is in `aea-config.yaml`

Before the recent tomte canonical-migration PR (#201), market-creator deployed cleanly on Propel. After #201 landed, the deployed container started crashing at startup with a missing-hypothesis import error. Root cause: the synced `abstract_round_abci` skill bundles `tests/__init__.py` in its fingerprint, which does `from hypothesis import settings` at module level. AEA imports the skill package on startup; the Propel base image (`open-autonomy[all]==0.21.20` — *not* `open-aea-test-autonomy`) does not ship hypothesis transitively, so without an explicit pin in the agent's `dependencies:` block, `autonomy build-image` produces a container that fails at boot.

The dev venv (and `make run-agent`) works either way because `uv sync --all-groups` installs hypothesis via `tomte` / `open-aea-test-autonomy` dev extras — local testing does **not** reproduce the Propel failure.

This PR adds the explicit pin to restore the working state. Proper upstream fix (drop the test bundle from the skill fingerprint or make the import lazy) is out of scope here — will be filed as a follow-up against open-autonomy.

## Test plan

- [x] Lint pipeline green locally (isort, black, pylint, mypy, flake8, darglint)
- [x] Unit tests green — 606 tests, 100% coverage
- [x] `test_get_claim_params_success` regression-locked: fails if `chunk_size` is removed from the contract call
- [x] `gql.gql('{ a { b } }')` parses successfully after the `graphql-core` pin (was raising `FrozenInstanceError` on the alpha)
- [x] CI green
- [x] Smoke-test on Propel after merge: confirm `Event.REALITIO_WITHDRAW_BONDS_TX_DONE` fires within one cycle, and the `propose-question` Mech tool runs without the AST mutation error

## Out of scope (follow-ups)

- Upstream issue against `open-autonomy`: drop `tests/__init__.py` from `abstract_round_abci`'s `skill.yaml` fingerprint (or make hypothesis import lazy) so consumers don't need the runtime pin.
- `mech-interact` v0.31.0 introduces an allowlist breaking change (replaces the `irrelevant_tools` blacklist). Not in this PR — current pin stays at v0.30.0. Will be handled in the next mech-interact bump alongside trader and market-resolver.


DhairyaPatel7 (dismissed):

packages/valory/agents/market_maker/aea-config.yaml
```diff
@@ -247,6 +250,6 @@ public_id: valory/http_server:0.22.0:bafybeicblltx7ha3ulthg7bzfccuqqyjmihhrvfezt
 type: connection
 config:
   host: ${str:0.0.0.0}
-  port: ${int:8716}
+  port: ${int:8000}
```
LOCKhart07, line 255: Not sure why this is being changed

LOCKhart07 (dismissed):

packages/valory/agents/market_maker/aea-config.yaml
jmoreira-valory: Because it's required for Propel - common source of issues if the environment variable is missed.

LOCKhart07: so propel already overrides it to 8000?

jmoreira-valory: No, Propel needs 8000 - but this is most times missed as an environment variable, thus the service is not rachable by the healthcheck server.

packages/valory/agents/market_maker/aea-config.yaml
```diff
@@ -92,6 +92,8 @@ dependencies:
     version: ==3.5.0
   requests-toolbelt:
     version: ==1.0.0
+  hypothesis:
```
LOCKhart07, line 97: `hypothesis==6.21.6` is being added to the agent's runtime `dependencies:` block — but `hypothesis` is a property-testing library, only imported under `tests/` modules. The trader repo's `aea-config.yaml` doesn't ship hypothesis as a runtime dep, and `open-aea-test-autonomy==0.21.20` doesn't pull it in transitively (per `uv.lock`). If something in `abstract_round_abci/tests/__init__.py` is being imported at agent startup and forcing this, a one-line comment here (or a note in the PR body) explaining why would help. Otherwise this is pulling a test-only lib into production deployments.

This change was also bundled into the `fix: ports` commit, which makes it easy to miss.

packages/valory/skills/omen_realitio_withdraw_bonds_abci/behaviours/withdraw_bonds.py
```diff
@@ -298,28 +298,30 @@ def _get_claim_params(
             from_block=from_block,
             to_block=to_block,
             question_id=question_id,
+            chunk_size=self.params.event_filtering_batch_size,
```
LOCKhart07, line 301: The existing `test_get_claim_params_success` test (`tests/behaviours/test_withdraw_bonds.py:76`) only patches `get_contract_api_response` and asserts on the returned 4-tuple — it doesn't verify that `chunk_size` is actually forwarded. A typo or upstream rename (e.g. trader renaming the param) would silently break chunking with the tests still green.

Suggest capturing the call kwargs and asserting `chunk_size == 1000` is forwarded — that's where the real coverage for this PR addition lives.

packages/valory/agents/market_maker/aea-config.yaml
LOCKhart07, line 255: Re the open follow-up on this thread ("so propel already overrides it to 8000?") — worth resolving before merge. If Propel sets `HTTP_SERVER_PORT=8000` via env var, the default change here is purely defensive. If Propel doesn't override and the port was simply mis-set to 8716, then this fix is what unblocks Propel. The distinction matters for whether other deployments need adjusting.

pyproject.toml
```diff
@@ -68,7 +68,7 @@ prerelease = "allow"
 # Realitio / FPMM contract wrappers and the mech_interact_abci skill.
 upstream_pins = [
     "valory-xyz/mech-interact@v0.30.0",
-    "valory-xyz/trader@v0.35.5",
+    "valory-xyz/trader@v0.37.0",
```
LOCKhart07, line 72: PR body says "fix: multisend, dependencies, ports" but the actual scope is broader:

- trader pin bump v0.35.5 → v0.37.0 (carries `simulate_claim_winnings` fix + bytes-by-construction contracts)
- new `event_filtering_batch_size` skill param to chunk `eth_getLogs` for Realitio claim params (rationale only in `models.py:62-65`)
- `hypothesis==6.21.6` added to agent runtime deps
- revert of the defensive `data.hex()` workarounds in 3 omen skills (now that contracts return bytes natively)

A couple of bullets in the description would make the diff much easier to follow for future readers / git-archaeology.

packages/valory/agents/market_maker/aea-config.yaml
jmoreira-valory: The change is corrective, not just defensive. Propel's k8s ingress is hard-wired to target port 8000 — Propel does **not** set `HTTP_SERVER_PORT=8000` automatically. With the old 8716 default, every Propel deployer had to remember to add `HTTP_SERVER_PORT=8000` to `.env.propel`, otherwise the ingress routed to a non-existent listener and the healthcheck endpoint silently returned 502.

Making 8000 the default means Propel deployments work out-of-the-box. Local `make run-agent` / custom infra that wants a different port can still override via the env var — backward-compatible.

For other repos still on 8716: anything hard-coding 8716 in infra should be audited; anything using env-var overrides is unaffected.

packages/valory/agents/market_maker/aea-config.yaml
jmoreira-valory: Good flag. Investigated: `hypothesis` IS required at runtime in the Propel container, despite being a test-only library upstream.

**Chronology:**
- Before the recent tomte canonical-migration PR (#201): market-creator deployed cleanly on Propel — `hypothesis` was available in the container.
- After #201 landed: market-creator started failing on Propel because `hypothesis` was no longer present in the container at runtime.
- This PR adds the explicit `hypothesis==6.21.6` pin to `aea-config.yaml dependencies:` to restore the pre-#201 working state.

**Why it's required at runtime, not just for tests:** the synced `abstract_round_abci` skill bundles `tests/__init__.py` in its `skill.yaml` fingerprint, and that file does `from hypothesis import settings` (guarded by `# pragma: nocover` but still evaluated at import time). When AEA loads the skill package during agent startup, the import fires. The Propel base image (`open-autonomy[all]==0.21.20` — *not* `open-aea-test-autonomy`) does not ship `hypothesis` transitively, so without an explicit pin in the agent's `dependencies:` block, `autonomy build-image` produces a container that crashes at startup.

The dev venv works either way because `uv sync --all-groups` installs hypothesis via `tomte` / `open-aea-test-autonomy` dev extras — that's why local `make run-agent` doesn't reproduce the failure. The Propel-clean-room behavior is the authoritative path.

**Proper upstream fix** (out of scope here): drop `tests/__init__.py` from `abstract_round_abci`'s `skill.yaml` fingerprint in open-autonomy, or make the hypothesis import lazy. Will file an issue against open-autonomy as follow-up so consumers don't need this pin on the next bump.

Updated the PR body to document the rationale so this doesn't get re-questioned on future framework upgrades.

packages/valory/skills/omen_realitio_withdraw_bonds_abci/behaviours/withdraw_bonds.py
jmoreira-valory: Done in `f247622`. Switched the patch from `make_gen(resp)` to `MagicMock(side_effect=make_gen(resp))` so `call_args` is tracked, added `context_mock.params.event_filtering_batch_size = 1000` to setUp, and asserted `mock_get.call_args.kwargs.get("chunk_size") == 1000`. Verified by stripping the kwarg from the behaviour and confirming the test now fails.

pyproject.toml
jmoreira-valory: Agreed, updated the PR body with the full summary covering: trader v0.35.5 → v0.37.0 pin bump (carries the `simulate_claim_winnings` fix + bytes-by-construction migration), the `event_filtering_batch_size` new param, the omen workaround reverts, the port default fix, and the test-tightening for chunk_size forwarding. Also documented the `hypothesis` runtime requirement (Propel container issue, see the response above) so it doesn't get re-questioned on the next framework bump.

LOCKhart07 (approved):

DhairyaPatel7 (approved):

PR #220, opened 2026-06-03
feat(propose_questions): verify extracted state sources are publicly resolvable

jmoreira-valory:
## Summary

The question-generator was emitting markets whose named sources (e.g. *"Tesla customer service repair records"*, *"Silicon Ranch operational records"*, *"NYC property tax records"* for billionaire counts) do not publicly publish the requested figure. These accumulate as `is_determinable=False` ("Case B") retries in the resolver, producing the ~50 pending markets piling up over the last weeks.

Two layered changes in `propose_questions.py`:

### 1. Self-review filter tightened from "3 of 4" to "all 4" soft checks

The Python filter was silently softer than the prompt: the prompt's step-I rule says *"Accept ONLY if ALL of B, C, D, E, F, G, H pass"* but the Python accepted on `passes >= 3`. This let questions through whose own `figure_is_directly_published` self-flag was `False`.

### 2. New `verify_state_is_resolvable()` step between state extraction and question generation

For each extracted `(source, metric)` tuple, the new function:
- Issues one Serper Google search restricted to relevant context
- Asks `gpt-4.1-mini` whether the top organic snippets show the source publishing the figure at a checkable cadence
- Drops the state if the judge says no

A simple Serper hit-count is too noisy — Google fuzzy-matches and gives 9-10 hits to clearly fictional source names. The LLM-on-snippets approach correctly distinguishes real sources (Freddie Mac PMMS, SEC 10-Q, US Census Bureau) from fictional ones (Tesla customer service records, Silicon Ranch operational records, CDC volunteer records).

If **all** states fail verification, the cycle is skipped (hard veto). Without this, the generator falls back to `"(none found)"` and tends to re-emit a question referencing the source the verifier just rejected, propagating the failure mode.

Fail-open on Serper / OpenAI errors — a transient outage never starves question generation.

## Validation (50-run local benchmark)

| Metric | Value |
|---|---|
| Hard-veto fire rate (all states rejected) | **4 %** (2 / 50 runs) |
| Average state drop rate | **27 %** (verifier rejects ~1 in 4 extracted states) |
| Verifier overhead | **~$0.005 / run** (4 Serper + 4 gpt-4.1-mini calls) |
| Per-run wall-clock | 42 s (was ~38 s without verifier) |
| All 287 unit tests | pass |

### Calibration on the 8-case fixture (real + fictional sources)

| Expected | Verdict |
|---|---|
| REAL: Freddie Mac PMMS | KEEP |
| REAL: SEC 10-Q filing | KEEP |
| REAL: US Census Bureau | KEEP |
| REAL: S&P 500 closing value | KEEP |
| **FICTIONAL: Tesla customer service repair records** | **DROP** |
| **FICTIONAL: Silicon Ranch operational records** | **DROP** |
| **FICTIONAL: CDC employee volunteer records** | **DROP** |
| **AUTHORITY_SILENT: Russian MoD troop count** | **DROP** |

11/13 correct including all dominant production failure patterns. Two known false positives (Eclipse X9 basalt fooled by tangential materials-science snippets, FiveThirtyEight stale-but-indexed) — acceptable as a soft filter.

## Test plan

- [ ] Lint pipeline: isort-check, black-check, flake8, mypy, pylint, darglint, check-abciapp-specs — all green locally
- [ ] All 287 unit tests pass
- [ ] `autonomy packages lock` updated (hash bumps cascaded to aea-config / service.yaml / parent skill)
- [ ] Manual: confirm new sources from the next service deployment do not include obviously-fictional ones


OjusWiZard (commented):
Multi-perspective review (correctness, silent failures, comments, types, tests, simplification). No critical issues; core logic is sound and the IPFS-hash cascade across skill/agent/service files is consistent. Three findings posted inline — the line 884 one is the only one I'd recommend addressing before merge.

packages/valory/skills/market_creation_manager_abci/propose_questions.py
```diff
@@ -777,6 +777,114 @@ def scrape_url(serper_api_key: str, url: str) -> Optional[dict]:
...
+        content = judge_response.choices[0].message.content or ""
+        out = json.loads(content)
+    except (
+        openai.OpenAIError,
+        json.JSONDecodeError,
+        AttributeError,
+        KeyError,
+        TypeError,
+    ):
+        return True, "fail_open_llm_error"
```
OjusWiZard, line 887: **Important — breaks the documented fail-open contract.** The `out.get(...)` extraction below is *outside* the `try/except`. If the judge returns valid JSON where `reason` is `null` (`None[:120]` → `TypeError`), or a body that parses to a non-dict (`out.get` → `AttributeError`), the exception escapes `verify_state_is_resolvable`, bubbles up to `with_key_rotation`'s blanket `except Exception`, and **aborts the entire run** — the opposite of the "a transient outage cannot starve generation" promise in the docstring. `response_format={"type":"json_object"}` guarantees `out` is a dict, but it does *not* guarantee `reason` is a string.

Suggest coercing (and/or moving inside the `try`):
```python
return out.get("answer") == "YES", str(out.get("reason") or "")[:120]
```

packages/valory/skills/market_creation_manager_abci/propose_questions.py
```diff
@@ -777,6 +777,114 @@ def scrape_url(serper_api_key: str, url: str) -> Optional[dict]:
...
+- YES only if a snippet clearly shows the source publishing this figure
+  (or a strongly equivalent figure on the same cadence).
+- NO if snippets are only news commentary, speculation, expert opinion,
+  or about a related-but-different topic.
+- NO if no snippet contains a specific number / value / measurement tied
+  to this metric.
+"""
+
+
+def verify_state_is_resolvable(
```
OjusWiZard, line 805: **Important — new gating logic has no test coverage.** This file is omitted in `.coveragerc`, so the repo's 100% coverage gate gives false assurance here. `verify_state_is_resolvable` (5 distinct fail-open/drop branches), the new hard-veto skip path, and the `passes >= 3` → `passes == len(checks)` flip all directly gate market creation. The function is a pure boundary-mocked unit (`requests.post` + `client.chat.completions.create`) — same shape already mocked for `scrape_url`/`gather_articles` — so the coverage exclusion isn't a hard blocker to testing it. At minimum the non-200, `RequestException`, no-organic→drop, and YES/NO branches deserve targeted tests. (Flagging as a judgment call if the file-level exclusion is intentional.)

packages/valory/skills/market_creation_manager_abci/propose_questions.py
```diff
@@ -777,6 +777,114 @@ def scrape_url(serper_api_key: str, url: str) -> Optional[dict]:
...
+    headers = {"X-API-KEY": serper_api_key, "Content-Type": "application/json"}
+    try:
+        response = requests.post(
+            "https://google.serper.dev/search",
+            headers=headers,
+            data=payload,
+            timeout=30,
+        )
+        if response.status_code != 200:
+            return True, f"fail_open_serper_status_{response.status_code}"
```
OjusWiZard, line 842: **Advisory.** A 500 returns `(True, fail_open_...)` (keep), but a clean 200 whose body lacks an `organic` key collapses to `[]` → `(False, "no_hits")` (drop). For genuinely fictional sources the drop is the intended signal, but it means "Serper succeeded with an unexpected shape" is indistinguishable from "Serper returned zero results" — the one fail-*closed* path in an otherwise fail-open function. Low impact; noting only because it slightly contradicts the stated philosophy.

jmoreira-valory:
All three review comments addressed:

**Comment 1 (line 884, Important)**: Fixed with `str(out.get("reason") or "")[:120]` -- coerces null to empty string before slicing. Prevents TypeError on null reason from LLM judge.

**Comment 2 (line 805, Important)**: Added `test_propose_questions.py` with 10 tests covering all `verify_state_is_resolvable` branches: empty query, Serper non-200, network error, missing organic key, empty organic, no client, judge YES, judge NO, null reason, LLM error. Hoisted the openai mock to parent `tests/conftest.py` with a proper `OpenAIError` exception base class so all test files share it.

**Comment 3 (line 842, Advisory)**: Now explicitly checks `"organic" not in body` and returns `(True, "fail_open_serper_unexpected_shape")` -- distinguishing missing key (fail-open) from genuinely empty results (fail-closed drop). The fail-open contract is now consistent across all Serper response paths.

packages/valory/skills/market_creation_manager_abci/tests/test_propose_questions.py
```diff
@@ -0,0 +1,159 @@
...
+
+from packages.valory.skills.market_creation_manager_abci.propose_questions import (
+    verify_state_is_resolvable,
+)
+
+_MODULE = "packages.valory.skills." "market_creation_manager_abci.propose_questions"
+
+
+class TestVerifyStateIsResolvable:
+    """Tests for verify_state_is_resolvable."""
```
OjusWiZard, line 34: Nice coverage of `verify_state_is_resolvable` — all 10 branches including the null-reason regression. Advisory only: the PR's other two behavior changes remain untested — the **hard-veto early-return** in `run()` (returns the error tuple when every state fails verification) and the **`passes >= 3` → `passes == len(checks)`** self-review threshold flip. Both live in the monolithic, coverage-excluded `run()`, so this is understandable; noting it for completeness since those are the actual control-flow changes that gate market creation.

packages/valory/skills/market_creation_manager_abci/tests/test_propose_questions.py
```diff
@@ -0,0 +1,159 @@
...
+
+import json
+import sys
+from unittest.mock import MagicMock, patch
+
+from packages.valory.skills.market_creation_manager_abci.propose_questions import (
+    verify_state_is_resolvable,
+)
+
+_MODULE = "packages.valory.skills." "market_creation_manager_abci.propose_questions"
```
OjusWiZard, line 30: Cosmetic: this uses implicit string concatenation (`"packages.valory.skills." "market_creation_manager_abci.propose_questions"`). The joined string is <88 chars, so the split is unnecessary and reads like an accidental missing `+`/`,`. Consider a single literal:
```python
_MODULE = "packages.valory.skills.market_creation_manager_abci.propose_questions"
```

OjusWiZard (dismissed):
Approving. The fix commit cleanly addresses all three findings from the first review, with a thorough new test suite for `verify_state_is_resolvable` (including a null-reason regression test). Two advisory notes inline — neither blocks. One heads-up not posted inline: `tests/behaviours/conftest.py:22` now has an unused `import sys` (the `sys.modules[...]` setup moved to the parent conftest) which fails `tox -e flake8` (F401) — verified against the tomte flake8 config; black/isort are clean. That one-liner should land before merge so CI goes green.

OjusWiZard (approved):

LOCKhart07 (approved):
Re-reviewed against a fresh `main`. All three earlier review points are addressed (null-reason coercion, the new `test_propose_questions.py` suite, and the `fail_open_serper_unexpected_shape` path). Core logic is sound: `verify_state_is_resolvable` is fail-open on every Serper/OpenAI error and fail-closed only on a clean 200 with zero hits; the `out.get(...)` line is safe under `response_format=json_object`; `passes == len(checks)` matches the documented 7-check rule; and the removed `(none found)` fallback is unreachable-by-design behind the hard veto. New tests are falsifiable and exercise real branching.

Two non-blocking observations (no need to gate the merge on these):

1. **Token/cost accounting gap.** The per-state judge calls in the Step 2b loop don't thread usage into `counter_callback`, unlike every other `client.chat.completions.create` in `run()` (e.g. the extraction call just above). So the ~$0.005/run verifier overhead is invisible to the mech tool's accounting. Likely intentional given the cheap `LIGHT_MODEL`, but flagging in case cost tracking matters.

2. **Nit:** when extraction yields zero states, the loop is a no-op and the hard veto fires with `"All extracted measurable states failed source verification"` / `states_proposed: 0`. The skip is correct, but the message attributes it to verification failure rather than "no states extracted."

LGTM.

PR #217, opened 2026-05-21
feat(config): restore + extend named env vars in aea-config.yaml

DhairyaPatel7:
## Summary

Restores the named `${VAR_NAME:type:default}` env-var syntax in `packages/valory/agents/market_maker/aea-config.yaml`, which was reverted in commit `b29c12f` ("fix: remove uppercase params from aea-config.yaml") because open-autonomy and the deployment middleware lacked support for the named form at the time. Both upstreams have since shipped that support, so the named form is now safe to re-introduce.

Also names a small number of new operator-tunable parameters that were added in anonymous form (`${type:default}`) after the original PR.

### Restored named env vars (83)

Re-applied the named form by reversing commit `b29c12f`. All previously-named vars are restored, e.g.:

- `LOG_FILE`, `LOG_LEVEL`
- `GNOSIS_LEDGER_RPC`, `GNOSIS_LEDGER_CHAIN_ID`, `GNOSIS_LEDGER_IS_POA_CHAIN`, `GNOSIS_LEDGER_PRICING`
- `ACN_URI`, `ACN_NODE_PUBLIC_KEY`
- `LOG_DIR`
- `MAX_PROPOSED_MARKETS`, `MAX_APPROVED_MARKETS`, `MARKETS_TO_APPROVE_PER_EPOCH`, `MARKETS_TO_APPROVE_PER_DAY`, `MIN_APPROVE_MARKETS_EPOCH_SECONDS`, `APPROVE_MARKET_EVENT_DAYS_OFFSET`
- `MARKET_APPROVAL_SERVER_URL`, `MARKET_APPROVAL_SERVER_API_KEY`
- `NEWSAPI_ENDPOINT`, `NEWSAPI_API_KEY`, `TOPICS`, `NEWS_SOURCES`
- `MARKET_FEE`, `INITIAL_FUNDS`, `MARKET_TIMEOUT`, `MIN_MARKET_PROPOSAL_INTERVAL_SECONDS`
- `EVENT_OFFSET_START_DAYS`, `EVENT_OFFSET_END_DAYS`, `MARKET_PROPOSAL_ROUND_TIMEOUT_SECONDS_PER_DAY`, `MAX_MARKETS_PER_STORY`
[...]

PR #221, opened 2026-06-04
fix(config): revert named runtime setup params on market_maker aea-config

DhairyaPatel7:
## Summary

Reverts the three agent setup-block runtime params named by #217 back to the anonymous form. They are runtime/consensus values that the deployment computes and injects — naming them breaks Pearl/`autonomy deploy` consensus.

```diff
-        all_participants: ${ALL_PARTICIPANTS:list:["0x0000000000000000000000000000000000000000"]}
-        safe_contract_address: ${SAFE_CONTRACT_ADDRESS:str:0x0000000000000000000000000000000000000000}
-        consensus_threshold: ${CONSENSUS_THRESHOLD:int:null}
+        all_participants: ${list:["0x0000000000000000000000000000000000000000"]}
+        safe_contract_address: ${str:0x0000000000000000000000000000000000000000}
+        consensus_threshold: ${int:null}
```

## Root cause

`ServiceBuilder.try_update_runtime_params` (`open-autonomy/autonomy/deploy/base.py`, and Pearl's middleware override) overwrites `safe_contract_address`, `all_participants` and `consensus_threshold` in the service overrides with the literal runtime-computed values **before** open-autonomy's `_extract_named_env_var_defaults` runs. Once the template string is replaced by a literal, there is no `${NAME:...}` placeholder left for the extractor to find, so no named key (`ALL_PARTICIPANTS`, `SAFE_CONTRACT_ADDRESS`, `CONSENSUS_THRESHOLD`) lands in `agent.json`.

open-aea then ignores the path-based fallback key (`SKILL_..._SETUP_*`) because the agent template is in named form, so the agent falls back to the zero-address / null YAML defaults in `aea-config.yaml`, and the consensus round rejects the real agent (`TransactionNotValidError: ... not in list of participants: ['0x0000...0000']`).

This is the same class of bug that optimus fixed in valory-xyz/optimus#330 (which reverted `all_participants`). market-creator named all three of these runtime setup fields, so all three are reverted here. `service.yaml` keeps the named form — it is unaffected, since `try_update_runtime_params` rewrites it anyway and the path-based key is what reaches the agent.

The durable fix lives in olas-operate-middleware / open-autonomy: teach `try_update_runtime_params` (or the extraction step) to also push the resolved value as a named key when it rewrites the override. Once that lands, these can return to the named form.

## Test plan

- [x] `tomte tox -e check-hash`, `check-packages`, `check-third-party-hashes`
- [x] `tomte tox -e check-dependencies`, `liccheck`, `check-doc-hashes`
- [x] `tomte tox -e check-abciapp-specs`, `check-abci-docstrings`, `check-handlers`, `analyse-service`
- [x] `tomte tox -p -e black-check -e isort-check -e flake8 -e mypy -e pylint -e darglint`
- [x] `tomte tox -p -e safety -e bandit`, `gitleaks`
- [x] `tomte tox -e py3.11-darwin` — 308 passed
- [x] `uv lock --check`, `uv sync --all-groups`
- [ ] Pearl / `autonomy deploy` smoke test: confirm `TransactionNotValidError` is gone and consensus proceeds with the real agent address

### DIFF
--- a/packages/packages.json
+++ b/packages/packages.json
@@ -3,8 +3,8 @@
         "contract/valory/fpmm_deterministic_factory/0.1.0": "bafybeiazddwzm6dc6625dkynx4devvfy5rq3fbh3vrpj4h66r7lusj6vua",
         "skill/valory/market_creation_manager_abci/0.1.0": "bafybeih33cswnqzlqtqf3hafcvipj3jappc7heobxox6ie4xgou73wx7yq",
         "skill/valory/market_maker_abci/0.1.0": "bafybeiawxosgijccnhlzpgd225vln6ntcerpkm4pdwlzxnsuv2jzhukcla",
-        "agent/valory/market_maker/0.1.0": "bafybeial2i2jboq6fdw7sq4dmdvrqck6pbocjt6xb5k7dzhemkaxjzhw44",
-        "service/valory/market_maker/0.1.0": "bafybeieukjhjgaptcpqnoatup72wm7fo7hct7bto26nr6ug3ynp7k3kzuy"
+        "agent/valory/market_maker/0.1.0": "bafybeifsn4o4gupgfc52qcza6arge5uhpsoiqethafpegxtw5r333buokq",
+        "service/valory/market_maker/0.1.0": "bafybeibeum27eb4zlfi5mopyt6a4psklzia45siigzxurdker2tywe3o7a"
     },
     "third_party": {
         "protocol/valory/contract_api/1.0.0": "bafybeibld2xb5m7kyluiptkamp4nrt6oeomkohz7a3yppbv2oo7qw2e4la",
--- a/packages/valory/agents/market_maker/aea-config.yaml
+++ b/packages/valory/agents/market_maker/aea-config.yaml
@@ -177,9 +177,9 @@ models:
       service_id: ${SERVICE_ID:str:market_maker}
       service_registry_address: ${SERVICE_REGISTRY_ADDRESS:str:0x9338b5153AE39BB89f50468E608eD9d764B755fD}
       setup:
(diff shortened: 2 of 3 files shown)


LOCKhart07 (approved):

jmoreira-valory (approved):
Tested locally on the same agent setup (PR #220 verifier branch + market-creator v0.12.3):

**Short-name form (broken behaviour PR #221 is fixing):**
- \`.env\` set with \`SAFE_CONTRACT_ADDRESS=0x98973c9c66a79B6cc7ed655D830D1AD84067c5E5\` and \`ALL_PARTICIPANTS='[\"0xce43622380…\"]'\` (the new named short form from #217)
- Log line 3003: \`Service owner identified: service_id=199, safe=0x0000000000000000000000000000000000000000, owner=0xeB9D8d…\`
- Subgraph query at line 3097: \`_collect_latest_open_markets for creator='0x0000000000000000000000000000000000000000'\`

Safe falls back to the YAML zero-address default exactly as the PR body predicts.

**Fully-qualified form (after PR #221):**
- \`.env\` switched to \`SKILL_MARKET_MAKER_ABCI_MODELS_PARAMS_ARGS_SETUP_SAFE_CONTRACT_ADDRESS=0x98973c9c…\` and the matching \`SKILL_..._SETUP_ALL_PARTICIPANTS=[…]\`
- Log line 3003: \`Service owner identified: service_id=199, safe=0x98973c9c66a79B6cc7ed655D830D1AD84067c5E5, owner=0xeB9D8d…\`
- Funds-forwarder (line 3008): real token balance / retain / transfer values used; transfer destination is the real owner address

Repros the bug + confirms the fix end-to-end. Approving.

PR #222, opened 2026-06-07
feat(propose_questions): soft veto + cache verifier results

jmoreira-valory:
## Summary

Tunes PR #220's `verify_state_is_resolvable` step to recover throughput on news days where the verifier rejects most candidates.

Observed problem (Pearl, June 7 production logs): 18 markets created over **8.95 h** instead of the expected ~3 h. The largest gaps (150 min, 88 min, 55 min) coincide with **stretches where every cycle hits the hard-veto** — sample window 03:55–06:25 UTC ran 14 FSM cycles, only 1 produced a market, with 13× `NO_MARKETS_RETRIEVED` and 2× re-verification of the same "Putin Vows to Boost Air Defences" article.

### 1. Soft fall-through on zero-verified states

`packages/valory/skills/market_creation_manager_abci/propose_questions.py:1184`

**Before**: if 0/N states verified → early-return the whole cycle with an error.

**After**: WARN log + `states_string = "(none found)"` and let the question-generator work from the article body alone. The 4-of-4 self-review gate downstream still rejects un-resolvable framings.

Rationale: the hard-veto was a safety net for the original `propose-question` failure mode (un-publishable sources reaching on-chain). The downstream self-review is a second safety net; we don't need both blocking together. Relaxing the upstream gate restores throughput on news days the verifier doesn't grok.

### 2. In-memory cache for `verify_state_is_resolvable`

Module-level dict keyed by `(source, metric)`. Bounded at 1024 entries with FIFO eviction. Fail-open outcomes (Serper HTTP error, LLM timeout, etc.) are **not cached** so transient outages can't poison the cache; only deterministic outcomes (YES / NO / no_hits) are stored.

On news days with topic-repeat (we saw the same Putin/Ukraine article re-verified within 150 min), the second call returns instantly without burning Serper + LLM. Saves cost and unblocks the FSM cycle.

## Tests

- 10 existing tests still pass — added a `pytest` autouse fixture in `tests/conftest.py` that clears `_VERIFY_CACHE` between tests so the cache doesn't leak across cases.
- 2 new tests:
  - `test_cache_hit_skips_serper_and_llm`: second call with same `(source, metric)` returns cached, no new HTTP/LLM calls
  - `test_fail_open_not_cached`: transient error → second call re-issues (verifies fail-open path isn't cached)

## Test plan

- [x] All 244 unit tests pass (`pytest packages/valory/skills/market_creation_manager_abci/tests/`)
- [x] Lint pipeline: isort-check, black-check, flake8, darglint all green
- [x] `autonomy packages lock` cascade applied
- [ ] Watch next 24h: market-creation cadence should tighten (target: median gap drops from ~22 min to ~12 min)

## Potential follow-ups (after observing PR-222 in production)

A forced-veto smoke test (10 cycles, 100% conflicting input) showed PR-222 produces 7/10 questions vs deployed's 0/10 — clear throughput recovery — but self-review accepted some borderline cases (e.g. "Will BA announce fare increase between Jun 8-13" — narrow window on a speculative announcement). Self-review's `figure_is_directly_published` check works, but `authority_can_act_in_time` is letting through some announcement-framed questions on 5-day windows that are unlikely to resolve cleanly.

If post-deploy monitoring shows an uptick in jury-undeterminable markets, two targeted follow-ups (in order of expected impact):

### Follow-up 1 — Tighten `authority_can_act_in_time` in `SELF_REVIEW_PROMPT`

Smallest possible change: prompt-engineering on the existing gate. The current self-review accepts 3/3 candidates on speculative announcements like "Will BMW or Porsche announce plans to close China facilities in 5 days?" — adding stricter language around "narrow-window corporate / governmental announcements" should catch these. No new infrastructure, no new code paths, just a stricter prompt on the existing heavy-model review step.

### Follow-up 2 — Bias `SELECT_STORY_PROMPT` harder against announcement-only articles

The existing prompt already says "AVOID articles whose only angle is 'will authority X announce Y?' unless a scheduled announcement is specifically anticipated". On news-light days the article picker still selects ceasefire / corporate-announcement stories. Strengthening this with explicit topic exclusions or a stronger continuation-friendly bias would reduce the upstream rate of articles that lead to weak announcement-framed questions in the first place.

Both follow-ups are prompt-only changes — no architectural shift. They should be evaluated only if production data warrants them; PR-222 alone may already produce an acceptable quality / throughput balance.

## Note on N = 0 path (extractor returned no states)

The old guard `if not verified_states:` fired for both N > 0 (states extracted, all dropped — the case this PR is fixing) and N = 0 (extractor returned nothing). The new `if not verified_states and states:` only fires the WARN for the N > 0 case. For N = 0, the function falls through to question generation with `states_string = "(none found)"` — matching the pre-PR-220 behavior (the `(none found)` sentinel and its associated prompt rule have existed since the iteration-2 commit on April 16). No behavior regression; the WARN is just less noisy.


packages/valory/skills/market_creation_manager_abci/propose_questions.py
```diff
@@ -1181,33 +1202,21 @@ def run(**kwargs: Any) -> Tuple[Optional[str], Optional[Dict[str, Any]], Any, An
...
+                # let the question-generator work from the article body
+                # alone -- the self-review pass downstream is the next
+                # gate that still rejects un-resolvable framings.
+                print(
+                    f"WARN: 0/{len(states)} states verified; falling back to "
+                    f"article-only generation (no measurable_states context)."
                 )
             states = verified_states
-            states_string = json.dumps(states, indent=2)
+            states_string = json.dumps(states, indent=2) if states else "(none found)"
```
OjusWiZard, line 1233: **Important — prompt rule mismatch**

The sentinel `"(none found)"` is placed under the `MEASURABLE_STATES` heading in the prompt, but the prompt rule at line 233 fires on `"If MEASURABLE_STATES is empty"` — and `"(none found)"` is not an empty section. In practice a capable model interprets it correctly, but the rule won't fire literally for a smaller/cheaper model variant. A one-word fix either here or in the prompt:

```python
# Option A — keep sentinel, update prompt rule (line 233):
# "If MEASURABLE_STATES shows (none found) or is empty, ..."

# Option B — use empty JSON array instead:
states_string = json.dumps(states, indent=2) if states else "[]"
```

Option B also has the advantage that the prompt rule fires on the literal empty-array string without any text edit.

packages/valory/skills/market_creation_manager_abci/propose_questions.py
```diff
@@ -1181,33 +1202,21 @@ def run(**kwargs: Any) -> Tuple[Optional[str], Optional[Dict[str, Any]], Any, An
...
-                            ),
-                            "tool": tool,
-                            "article_title": article.get("title", ""),
-                            "states_proposed": len(states),
-                        }
-                    ),
-                    None,
-                    None,
-                    counter_callback,
+            if not verified_states and states:
```
OjusWiZard, line 1219: **Minor — silent N=0 behavior change not called out in the PR description**

The old `if not verified_states:` hard-returned for two distinct cases:
- **N > 0**: states extracted but all dropped (the case this PR is fixing)
- **N = 0**: the LLM extracted nothing at all (`states = []`)

The new `and states` guard suppresses the WARN for N=0 but does NOT restore the early return — when `states = []`, the function now falls through to question generation with `states_string = "(none found)"`. This is actually a good change (an article with no measurable states can still produce an announcement question, which the prompt already supports at line 233), but the PR description only mentions N>0. Worth adding a sentence to the PR body so future readers do not have to rediscover this.

packages/valory/skills/market_creation_manager_abci/tests/test_propose_questions.py
```diff
@@ -157,3 +157,43 @@ def test_judge_llm_error_fails_open(
...
+        mock_client.chat.completions.create.return_value.choices = [choice]
+        ok1, reason1 = verify_state_is_resolvable("k", "Freddie Mac", "mortgage rate")
+        ok2, reason2 = verify_state_is_resolvable("k", "Freddie Mac", "mortgage rate")
+        assert ok1 is True and ok2 is True
+        assert "[cached]" in reason2
+        # Cache hit must NOT re-issue Serper or LLM call
+        assert mock_post.call_count == 1
+        assert mock_client.chat.completions.create.call_count == 1
+
+    @patch(f"{_MODULE}.client")
```
OjusWiZard: **Minor — unused mock**

`@patch(f"{_MODULE}.client")` injects `mock_client`, but the test throws at the `requests.post` level and never reaches the LLM path, so `mock_client` is dead weight. Either remove the decorator + parameter:

```python
@patch(f"{_MODULE}.requests.post")
def test_fail_open_not_cached(self, mock_post: MagicMock) -> None:
```

or, if you prefer to be defensive, add an explicit assertion that documents the intent:

```python
assert mock_client.chat.completions.create.call_count == 0
```

packages/valory/skills/market_creation_manager_abci/tests/test_propose_questions.py
```diff
@@ -157,3 +157,43 @@ def test_judge_llm_error_fails_open(
...
+            "organic": [{"title": "PMMS", "snippet": "6.5%"}]
+        }
+        body = {"answer": "YES", "reason": "publishes weekly"}
+        choice = MagicMock()
+        choice.message.content = json.dumps(body)
+        mock_client.chat.completions.create.return_value.choices = [choice]
+        ok1, reason1 = verify_state_is_resolvable("k", "Freddie Mac", "mortgage rate")
+        ok2, reason2 = verify_state_is_resolvable("k", "Freddie Mac", "mortgage rate")
+        assert ok1 is True and ok2 is True
+        assert "[cached]" in reason2
```
OjusWiZard, line 181: **Advisory — missing complementary assertion on first call**

The test proves the second call is cached but does not explicitly prove the first call was NOT cached. Adding the negative assertion makes the test self-documenting:

```python
assert "[cached]" not in reason1  # first call must be a cache miss
assert "[cached]" in reason2       # second call must be a cache hit
```

Without it, if the `_clear_verify_cache` fixture ever fails to clear between tests, `reason1` could also contain `[cached]` and the test would still pass.

OjusWiZard (commented):
Two surgical changes, well-executed. Four inline notes below — one important, three minor/advisory.

packages/valory/skills/market_creation_manager_abci/propose_questions.py
```diff
@@ -885,7 +902,11 @@ def verify_state_is_resolvable(
         TypeError,
     ):
         return True, "fail_open_llm_error"
-    return out.get("answer") == "YES", str(out.get("reason") or "")[:120]
+    result = (out.get("answer") == "YES", str(out.get("reason") or "")[:120])
+    if len(_VERIFY_CACHE) >= _VERIFY_CACHE_MAX:
+        _VERIFY_CACHE.pop(next(iter(_VERIFY_CACHE)))
+    _VERIFY_CACHE[cache_key] = result
```
LOCKhart07: **Nit + advisory** — this 3-line evict-and-insert block is duplicated verbatim above at the `no_hits` path (lines 865-867).

1. **DRY:** extract a small helper so the bound/eviction policy can't drift between the two call sites:
```python
def _cache_put(key: Tuple[str, str], result: Tuple[bool, str]) -> None:
    if len(_VERIFY_CACHE) >= _VERIFY_CACHE_MAX:
        _VERIFY_CACHE.pop(next(iter(_VERIFY_CACHE)), None)
    _VERIFY_CACHE[key] = result
```

2. **Eviction race (low confidence — depends on the Mech execution model):** `pop(next(iter(_VERIFY_CACHE)))` is a non-atomic check-then-pop. If `run()` is ever invoked concurrently in-process, two threads can read the same oldest key via `next(iter(...))` and the second `.pop(key)` raises an uncaught `KeyError` that fails the tool. Passing a default — `.pop(next(iter(_VERIFY_CACHE)), None)` (as in the helper above) — is a zero-cost hardening regardless of whether the race is actually reachable. If each `run()` is an isolated subprocess this is moot (but so would the cache's cross-cycle benefit be).

OjusWiZard (approved):
All raised findings addressed. One advisory thread (N=0 behaviour change) left open for the author to acknowledge. Good to merge.

LOCKhart07 (approved):
All review comments addressed in `cfd93c08`:

- Prompt rule now recognizes the `(none found)` sentinel (resolves the MEASURABLE_STATES mismatch).
- N=0 fall-through documented in the PR description.
- Unused `mock_client` removed; cache-hit test gains the `"[cached]" not in reason1` negative assertion.
- Eviction duplication extracted into `_cache_put()` with race-safe `pop(oldest, None)`.

Lock cascade re-run and hashes consistent across `packages.json`, `aea-config.yaml`, and `market_maker_abci/skill.yaml`. LGTM.

packages/valory/skills/market_creation_manager_abci/propose_questions.py
jmoreira-valory: Fixed in cfd93c08 — prompt rule at line 233 now reads *"If MEASURABLE_STATES is empty or shows the sentinel `(none found)`, you may create an announcement-style question..."* so smaller / cheaper model variants can match the rule literally. Kept the sentinel string itself for backwards-compat with the existing v2 (extractor returned no states) code path.

packages/valory/skills/market_creation_manager_abci/propose_questions.py
jmoreira-valory: Acknowledged — added a *"Note on N=0 path"* section to the PR description. The new `and states` guard only suppresses the WARN for the N=0 path (extractor returned nothing). The function still falls through to question generation with `states_string = "(none found)"`, which matches the pre-PR-220 behaviour (the `(none found)` sentinel and its prompt rule have existed since the iteration-2 commit on April 16). No behaviour regression — just a less noisy WARN.

packages/valory/skills/market_creation_manager_abci/tests/test_propose_questions.py
jmoreira-valory: Fixed in cfd93c08 — dropped the `@patch(f"{_MODULE}.client")` decorator + the `mock_client` param. Test now only patches `requests.post` (the fail-open path short-circuits before reaching the LLM, so the client mock was dead weight). Added a darglint docstring for the remaining `mock_post` param.

packages/valory/skills/market_creation_manager_abci/tests/test_propose_questions.py
jmoreira-valory: Added in cfd93c08 — both assertions are now explicit:

```python
assert "[cached]" not in reason1  # first call must be a cache miss
assert "[cached]" in reason2       # second call must be a cache hit
```

Catches any future regression in the `_clear_verify_cache` autouse fixture as you flagged.

packages/valory/skills/market_creation_manager_abci/propose_questions.py
jmoreira-valory: Fixed in cfd93c08 — extracted `_cache_put(key, result)` helper at module top; both call sites (`no_hits` path and judge-result path) now go through it. Used `pop(key, None)` with the default to defend against the eviction-after-clear race you flagged — agree it is low confidence given the Mech tool is single-threaded inside the agent process today, but the default is free defense and matches your suggestion verbatim.

PR #223, opened 2026-06-13
fix(approve_markets): run mech tool off the event loop to avoid FSM wedge

jmoreira-valory:
## Summary

Both market-creator instances (QS + Pearl) wedged completely -- FSM frozen 36-50h, zero round transitions -- while the `/healthcheck` endpoint kept reporting healthy. Root cause: `ApproveMarketsBehaviour` runs the question-generation mech tool **inline and synchronously** on the single AEA asyncio event loop.

```python
# approve_markets.py  # TODO THIS EMULATES MECH INTERACT SENDING REQUEST TO A TOOL
mech_tool_output = mech_tool_propose_questions.run(**tool_kwargs)[0]   # ~60-90s, blocking
```

While `run()` blocks (select -> scrape -> extract -> verify N states via Serper+LLM -> generate -> self-review), the loop cannot service the local Tendermint `begin_block` stream (nor inbound HTTP -- "Request timed out!" fires during the window). After ~60s the block-stall deadline expires, the framework attempts an emergency Tendermint hard-reset, and the FSM never resumes ("Period end." never logs again).

> **Cross-log confirmation (2026-06-16 Pearl recurrence).** The wedge recurred and was captured end-to-end across **both** the agent and tendermint container logs: Tendermint committed cleanly up to height 321, then stalled at 322 for ~63s waiting on the blocked ABCI app (the gap between `finalizing commit` and `executed block`), after which the agent's begin_block-deadline recovery hard-reset Tendermint to genesis (`Reset private validator file to genesis state`) and the FSM froze at period 17. Tendermint itself never crashed -- it resumed producing empty blocks from height 1 while the agent stayed silent (classic ABCI desync). The newly-deployed healthcheck (#224) caught it live: `is_healthy: false`, `liveness.reason: "no-fsm-data"`.

## Fix (minimal, surgical)

1. **Offload `run()` to a worker thread and yield.** Submit to a single-worker `ThreadPoolExecutor` and `yield from self.wait_for_condition(future.done)` so control returns to the agent loop each tick -- begin_block keeps being serviced while the tool runs. Same executor pattern `trader` already uses for blocking work.
2. **De-block the pacing sleeps.** Replace the three `time.sleep(3)` in `_propose_and_approve_market` with `yield from self.sleep(3)`; drop the now-unused `import time`.
3. **Harden the offload.** Degrade a raw tool exception to the existing `{"error": ...}` path (so a tool failure can't propagate and stall the skill), and `executor.shutdown(wait=False)` in `finally` -- never `wait=True`, which on `GeneratorExit` (service shutdown mid-run) would re-block the loop until the long call finished, the very thing this fix avoids.

No behavioural/logic change to question generation itself -- only how it is scheduled relative to the event loop.

## Diff

`approve_markets.py` +33/-9: the executor block (offload + exception guard + detached shutdown), the 3 one-line `self.sleep` swaps, and the import changes. Plus the `autonomy packages lock` cascade (dev packages only; third-party verified byte-identical to main).

## Test plan

- [x] `approve_markets` behaviour tests pass (the `_propose_and_approve_market_*` tests patch the behaviour `sleep` instead of `time.sleep`; suite has no real delays, confirming non-blocking) -- tool-exception path covered (100% branch coverage enforced)
- [x] isort / black / flake8 / darglint clean
- [x] `autonomy packages lock` cascaded (market_creation_manager_abci, market_maker_abci, agent, service); third-party byte-identical to main
- [x] CI green on the merge commit across ubuntu / macos / windows x py3.10-3.14
- [ ] Post-deploy: confirm no recurrence of the begin_block-deadline -> reset -> wedge sequence during `approve_markets_round`

## Context

Full root-cause investigation (two-layer model: this local trigger + a framework no-timeout reset hang) is in `.claude/plans/healthcheck_wedge_investigation.md`. The three-dimension `is_healthy` healthcheck model (#224) -- which reports a frozen FSM as unhealthy -- has since **merged to main and is included in this branch** (main merged in). This PR removes the *trigger*; #224 stopped the *lying healthcheck*. The framework-side reset-timeout hardening remains a separate upstream open-autonomy change.


packages/valory/skills/market_creation_manager_abci/behaviours/approve_markets.py
```diff
@@ -140,8 +140,29 @@ def _sender_act(self) -> Generator:
...
+                # The mech tool runs a long (~60-90s) synchronous LLM/HTTP
+                # pipeline. Running it inline blocks the single AEA event loop,
+                # which then stops servicing the local Tendermint `begin_block`
+                # stream; once the block-stall deadline expires the FSM wedges
+                # permanently. Offload to a worker thread and yield control back
+                # to the loop while it runs so consensus keeps progressing.
+                executor = concurrent.futures.ThreadPoolExecutor(max_workers=1)
+                future = executor.submit(mech_tool_propose_questions.run, **tool_kwargs)
+                try:
+                    yield from self.wait_for_condition(future.done)
```
LOCKhart07, line 152: Non-blocking: `wait_for_condition(future.done)` has no `timeout=`, so if the mech tool's internal HTTP/LLM hangs indefinitely (never returns, never raises) the future never completes and the `approve_markets` round stalls forever. This is strictly better than the old wedge — the loop keeps yielding each tick, so no `begin_block`-deadline hard-reset — and #224's healthcheck would now report the stalled FSM as unhealthy and trigger a restart. Since `wait_for_condition` does accept a `timeout=`, a reader will wonder why it's omitted here; worth a one-line note in the comment that the hang case is intentionally delegated to healthcheck-driven restart rather than a local timeout.

LOCKhart07 (approved):

packages/valory/skills/market_creation_manager_abci/behaviours/approve_markets.py
```diff
@@ -140,8 +140,29 @@ def _sender_act(self) -> Generator:
...
+                    mech_tool_output = future.result()[0]  # type: ignore
```
OjusWiZard, line 153: `run()` returns `Tuple[Optional[str], ...]`, so `[0]` can legitimately be `None`. Then `json.loads(None)` (next line) raises `TypeError`, which the broad `except` catches and logs as `"...the JSON object must be str... not NoneType"` — a symptom, not the cause. Same for a non-JSON tool string (`JSONDecodeError` logged generically, raw output never shown). It degrades *safely* to empty markets, so this is diagnosability, not a crash. Suggest:

```python
raw = future.result()[0]
if raw is None:
    self.context.logger.error("Mech tool returned None output (internal tool failure, no message)")
    mech_tool_output_json = {"error": "tool returned None"}
else:
    mech_tool_output_json = json.loads(raw)
```

packages/valory/skills/market_creation_manager_abci/behaviours/approve_markets.py
OjusWiZard, line 152: `wait_for_condition` with `timeout=None` busy-yields forever. This is backstopped by `ApproveMarketsRound`'s `ROUND_TIMEOUT = 180.0s` (on round change the behaviour is closed → `finally` detaches the thread), so it does **not** re-create a permanent wedge. But a hung tool currently routes through the round-timeout path instead of degrading cleanly. Passing an explicit `timeout` (< 180s) caught as `TimeoutException` → `{"error": "timeout"}` would let the keeper reach consensus normally. Optional.

packages/valory/skills/market_creation_manager_abci/behaviours/approve_markets.py
```diff
@@ -140,8 +140,29 @@ def _sender_act(self) -> Generator:
...
+                    mech_tool_output_json = json.loads(mech_tool_output)
+                except Exception as e:  # noqa  pylint: disable=broad-except
+                    # Degrade a raw tool exception to the existing tool-error
+                    # path instead of letting it propagate and stall the skill.
+                    self.context.logger.error(f"Mech tool raised an exception: {e}")
+                    mech_tool_output_json = {"error": str(e)}
```
OjusWiZard, line 159: Non-blocking design note: a persistent tool outage (bad API key, sustained provider failure) now silently lands here every cycle → `proposed_markets = {}` → exits via the "done" path. It logs at ERROR but never escalates, so the market backlog grows invisibly. Consider a consecutive-failure counter / distinct event for escalation. Out of scope for this fix — worth a tracking issue.

packages/valory/skills/market_creation_manager_abci/tests/behaviours/test_approve_markets.py
```diff
@@ -363,6 +363,51 @@ def test_sender_act_with_top_level_mech_error(self) -> None:
...
+                    self.behaviour, "send_a2a_transaction", new=_make_gen(None)
+                ),
+                patch.object(
+                    self.behaviour, "wait_until_round_end", new=_make_gen(None)
+                ),
+                patch.object(self.behaviour, "set_done") as mock_set_done,
+            ):
+                gen = self.behaviour._sender_act()
+                _exhaust_gen(gen)
+                mock_set_done.assert_called_once()
```
OjusWiZard, line 409: This test only asserts `set_done` was called, so the exception branch could be silently bypassed and it would still pass. More importantly, **no test asserts `executor.shutdown(wait=False)`** — the central design decision of this PR. If someone flips it to `wait=True` (the regression this PR exists to prevent), nothing fails. Suggest patching `ThreadPoolExecutor` to inject a mock and asserting `instance.shutdown.assert_called_once_with(wait=False)` plus `logger.error.assert_called()` here. (Coverage is 100%, but coverage ≠ behavioral assertion.)

OjusWiZard (approved):
Approving — the core fix is sound and correctly designed. Offloading the ~60-90s tool to a worker thread and `yield from self.wait_for_condition(future.done)` is the right way to keep the AEA loop servicing Tendermint, and `executor.shutdown(wait=False)` in `finally` is the correct choice (a `with` block would call `shutdown(wait=True)` and re-introduce the wedge on `GeneratorExit` — verified). `time.sleep -> self.sleep` swaps and the test updates are correct.

Leaving a few non-blocking inline notes. Two more out-of-scope items not worth inlining:

- **Pre-existing (not this PR):** `approved_markets_count += 1` (~line 185) is unconditional even when `_propose_and_approve_market` returns `ERROR_PAYLOAD` (return value discarded), so it over-counts on HTTP failures. Worth a separate fix.
- **Nits:** `tool_kwargs` is a single-use dict that could be inlined into `executor.submit(...)`; the `# TODO ... EMULATES MECH INTERACT` / `# END ... EMULATION` markers now bracket a hardened concurrency block and read as stale.

jmoreira-valory:
Thanks for the review. The non-blocking notes here — guarding a `None` tool output for clearer diagnostics, an explicit test assertion on `shutdown(wait=False)`, the doc note that the no-timeout hang is delegated to healthcheck-driven restart, and a consecutive-failure/escalation counter for persistent tool outages — are captured for the follow-up that moves question-generation onto the mech via `mech_interact_abci`. That work removes this inline `run()` block entirely, so the suggestions are tracked against the new request/response path rather than hardening code about to be deleted. Resolving these to merge the stopgap.

PR #224, opened 2026-06-13
fix(healthcheck): report is_healthy=false when FSM is frozen (mech parity)

jmoreira-valory:
## Summary

When both market-creator instances wedged (FSM frozen 36-50h), `/healthcheck` kept returning HTTP 200 with `is_tm_healthy: true` -- a healthy report for a dead agent. Two bugs in `market_maker_abci/handlers.py` `_handle_get_health`:

1. The `if round_sequence._last_round_transition_timestamp:` guard skips all liveness computation when the FSM is frozen, leaving `is_tm_unhealthy = None`.
2. `"is_tm_healthy": not is_tm_unhealthy` then turns that `None` into `True` (`not None == True`) -- so the deadest state reports the most confident `true`. There was also no single `is_healthy` boolean for a supervisor to key on.

This ports the mech service's three-dimension health model so a frozen FSM is correctly reported as `is_healthy: false`, converging the healthcheck shape with the mech fleet.

## Changes (surgical -- only `_handle_get_health` + constants)

- **Liveness** (the dimension that catches the wedge): `seconds_since_last_transition is None or is_tm_unhealthy is None` -> `(False, "no-fsm-data")`; `is_tm_unhealthy` -> `tm-unhealthy`; `seconds > LIVENESS_STALL_FACTOR * reset_pause` -> `stuck-no-transition`; else `ok`.
- **Readiness / Progress**: market-creator is FSM-only with no task backlog (unlike mech), so both are always-ok `idle-ok` placeholders kept for shape parity. No backlog tracking invented; mech's task-specific fields (`PENDING_TASKS`, `last_tx`, etc.) are deliberately NOT copied.
- `is_healthy = bool(liveness_ok and readiness_ok and progress_ok)`.
- Fix the not-None bug: `"is_tm_healthy": (None if is_tm_unhealthy is None else not is_tm_unhealthy)` (three-state).
- Constants mirror mech: `LIVENESS_STALL_FACTOR=3.0`, `TRANSITION_TOLERANCE_FACTOR=2.0`, `HEALTH_VERSION=2`.
- **HTTP 200 always** (`_send_ok_response`). No status-code change, no auto-restart -- the wedge is signalled solely via `is_healthy: false` (operator decision). `HttpCode` enum and all other handlers untouched.
- All existing observational fields preserved; added `is_healthy`, `liveness`, `readiness`, `progress`, `current_round`, `health_version`.

## JSON emitted for the wedged state (the bug being fixed)

```json
{
  "is_healthy": false,
  "liveness": {"ok": false, "reason": "no-fsm-data"},
  "readiness": {"ok": true, "reason": "idle-ok"},
  "progress": {"ok": true, "reason": "idle-ok"},
  "seconds_since_last_transition": null,
  "is_tm_healthy": null,
  "period": 2,
  "reset_pause_duration": 600.0,
  "current_round": null,
  "rounds": null,
  "is_transitioning_fast": null,
  "health_version": 2
}
```
Previously this exact state returned `is_tm_healthy: true` and no `is_healthy` field.

## Test plan

- [x] 36 tests pass (`test_handlers.py`), incl. 4 new health-dimension cases: recent->healthy, tm-unhealthy, stale->stuck-no-transition, no-last-transition->unhealthy; every case asserts HTTP 200, `readiness`/`progress == idle-ok`, `health_version == 2`
- [x] isort / black / flake8 / darglint clean
- [x] `autonomy packages lock` cascaded (dev packages only: market_maker_abci skill/agent/service; no third-party moved)

## Context

Companion to the wedge-prevention PR (de-blocks the `approve_markets` mech call that triggers the freeze). This PR stops the healthcheck from lying about it; that PR removes the trigger. Full root-cause: `.claude/plans/healthcheck_wedge_investigation.md`.

### DIFF
--- a/packages/valory/skills/market_maker_abci/handlers.py
+++ b/packages/valory/skills/market_maker_abci/handlers.py
@@ -17,7 +17,7 @@
 #
 # ------------------------------------------------------------------------------
 
-"""This module contains the handler for the 'price_estimation_abci' skill."""
+"""This module contains the handler for the 'market_maker_abci' skill."""
 
 import json
 import re
@@ -70,7 +70,6 @@ IpfsHandler = BaseIpfsHandler
 # Health thresholds (mirrored from the mech service for cross-service parity).
 LIVENESS_STALL_FACTOR = 3.0  # > 3x the expected pause means the FSM is "stuck"
 TRANSITION_TOLERANCE_FACTOR = 2.0  # < 2x the expected pause means "transitioning fast"
-DEFAULT_GRACE = 120.0  # readiness/progress grace floor, kept for mech parity
 HEALTH_VERSION = 2
 
 
--- a/packages/valory/skills/market_maker_abci/handlers.py
+++ b/packages/valory/skills/market_maker_abci/handlers.py
@@ -69,7 +69,9 @@ IpfsHandler = BaseIpfsHandler
 
 # Health thresholds (mirrored from the mech service for cross-service parity).
 LIVENESS_STALL_FACTOR = 3.0  # > 3x the expected pause means the FSM is "stuck"
-TRANSITION_TOLERANCE_FACTOR = 2.0  # < 2x the expected pause means "transitioning fast"
+TRANSITION_TOLERANCE_FACTOR = (
+    2.0  # < 2x the expected pause (and tm healthy) means "transitioning fast"
+)
 HEALTH_VERSION = 2
 
 
@@ -264,18 +266,16 @@ class HttpHandler(BaseHttpHandler):
         """
         Handle GET /healthcheck and compute the agent's health metrics.
 
-        The health is assessed across three dimensions (mech-service parity):
+        Only **liveness** is actually measured: whether the FSM is transitioning
+        and Tendermint is not stalled. A missing last-transition timestamp is
+        reported as not-live (reason ``no-fsm-data``) so a frozen FSM cannot
+        read as healthy. ``readiness`` and ``progress`` are always-ok
+        placeholders kept for shape parity with the mech service -- market-creator
+        is FSM-only with no task backlog or dependency-read metric.
 
-        - **Liveness**: whether the FSM is actively transitioning and Tendermint
-          is not in a stalled state. This is the dimension that catches a frozen
-          FSM: a missing last-transition timestamp is reported as not-live
-          (reason ``no-fsm-data``) instead of being silently treated as healthy.
-        - **Readiness**: whether the agent can accept new work.
-        - **Progress**: whether the agent is advancing through its backlog.
-
-        market-creator is an FSM-only service with no task backlog or dependency
-        read metric (unlike mech), so readiness and progress are always-ok
-        placeholders (reason ``idle-ok``) kept for shape parity with mech.
+        TODO: if this service later gains external I/O that can stall a
+        behaviour while the FSM still ticks via timeout transitions, compute
+        readiness/progress for real instead of these placeholders.
 
         The endpoint always responds with HTTP 200; a wedge is signalled solely
         via ``is_healthy: false`` in the JSON body.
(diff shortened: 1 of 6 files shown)


LOCKhart07 (dismissed):
Traced every code path and test — the fix is correct and the tests are falsifiable (they assert on real computed output, not the mock). The `not is_tm_unhealthy → True` coercion bug is genuinely fixed (now three-state), both wedge shapes are caught (`no-fsm-data` for a null timestamp, `stuck-no-transition` for a stale one, `tm-unhealthy` for a stalled TM), and the lock cascade across packages.json / aea-config / service.yaml / skill.yaml is the correct chain with no third-party movement. No in-repo consumer depends on the old health JSON shape.

Approving. Two non-blocking notes:

1. **Pre-first-transition reports `is_healthy: false` (`no-fsm-data`).** Before the FSM's first round transition (fresh startup / post-restart, while registration + Tendermint come up), `_last_round_transition_timestamp` is falsy, so the agent reads as unhealthy. Fine here since the endpoint always returns HTTP 200 (status-code-based liveness probes won't restart-loop) and `is_healthy` is an operator signal with no auto-restart — just worth confirming operator tooling tolerates a transient `false` at boot rather than alerting on it.

2. **Nit:** the `or` in `seconds_since_last_transition is None or is_tm_unhealthy is None` is redundant — both are set together inside Guard A, so they're always both-None or both-set. Harmless defensive code; no change needed.

packages/valory/skills/market_maker_abci/handlers.py
```diff
@@ -280,26 +303,53 @@ def _handle_get_health(
...
+        # Guard B: round identifiers for observability.
         if round_sequence._abci_app:
             current_round = round_sequence._abci_app.current_round.round_id
             rounds = [
                 r.round_id for r in round_sequence._abci_app._previous_rounds[-25:]
             ]
             rounds.append(current_round)
 
+        # Liveness: catches the wedge. No FSM data or unknown tm state -> not live.
+        if seconds_since_last_transition is None or is_tm_unhealthy is None:
```
OjusWiZard: **Important — dead operand.** Both `seconds_since_last_transition` and `is_tm_unhealthy` are assigned together, exclusively inside Guard A (`if round_sequence._last_round_transition_timestamp:`). There is no path that sets one without the other, so `is_tm_unhealthy is None` is true *iff* `seconds_since_last_transition is None` — the second operand never independently fires. The comment above ("No FSM data or unknown tm state") also implies two reachable triggers, but `live_reason` collapses both to `no-fsm-data`.

Either simplify to:
```python
if seconds_since_last_transition is None:
```
(clearer — it's the direct "no transition yet" proxy), or, if the `or` is intentional future-proofing for the two being decoupled later, say so in the comment. Behavior-preserving either way.

packages/valory/skills/market_maker_abci/handlers.py
```diff
@@ -280,26 +303,53 @@ def _handle_get_health(
...
+            liveness_ok, live_reason = False, "no-fsm-data"
+        else:
+            liveness_ok = (not is_tm_unhealthy) and (
+                seconds_since_last_transition <= LIVENESS_STALL_FACTOR * reset_pause
```
OjusWiZard, line 326: **Advisory — `reset_pause = 0` footgun.** If `reset_pause_duration` is ever configured/overridden to `0`, `LIVENESS_STALL_FACTOR * reset_pause == 0`, so `seconds_since_last_transition <= 0` is effectively never true and every poll reports `stuck-no-transition` even for a healthy FSM. Not reachable with the default (600/1800) and tests pin it to 60, so this is just optional hardening — a guard or `logger.warning` when `reset_pause <= 0` would make the misconfiguration visible.

packages/valory/skills/market_maker_abci/handlers.py
```diff
@@ -280,26 +303,53 @@ def _handle_get_health(
...
+            live_reason = (
+                "ok"
+                if liveness_ok
+                else ("tm-unhealthy" if is_tm_unhealthy else "stuck-no-transition")
+            )
+
+        # Readiness/progress: market-creator has no backlog metric, so these are
+        # always-ok placeholders kept for mech parity.
+        readiness_ok, ready_reason = True, "idle-ok"
+        progress_ok, prog_reason = True, "idle-ok"
```
OjusWiZard, line 338: **Advisory — placeholder framing.** `readiness`/`progress` are hardcoded `True`/`idle-ok`. Defensible for an FSM-only service, and the docstring does eventually call them placeholders — but the docstring leads by describing them as if measured ("can accept new work" / "advancing through its backlog") and only walks it back two paragraphs later, which is easy to miss. Consider stating up front that only Liveness is computed.

Also note the latent limitation: if the service later gains external I/O that stalls a behaviour while the FSM still ticks via timeout transitions, these two dimensions would lie. A `# TODO` marking the follow-up would help.

packages/valory/skills/market_maker_abci/handlers.py
```diff
@@ -280,26 +303,53 @@ def _handle_get_health(
...
+
+        is_healthy = bool(liveness_ok and readiness_ok and progress_ok)
```
OjusWiZard, line 340: **Advisory (nit) — redundant `bool()`.** `liveness_ok` is already `bool` and `readiness_ok`/`progress_ok` are `True` literals, so the `and` chain is already `bool`; the wrapping `bool(...)` is a no-op. `is_healthy = liveness_ok and readiness_ok and progress_ok` reads the same.

OjusWiZard (commented):
Automated multi-perspective review (code / tests / silent-failure / comments / simplify). CI is fully green, including the 100%-coverage matrix, and the core fix (three-state `is_tm_healthy` + frozen-FSM liveness) is correct and well-tested. Inline notes below are one cleanup nit plus minor advisories — none block merge.

One extra design note that doesn't map to a single line: the endpoint always returns HTTP 200, signalling the wedge only via `is_healthy: false`. That's a deliberate, documented choice, but any supervisor/probe that keys on the HTTP status code alone (rather than parsing the JSON body) will never see a wedged agent. Worth confirming downstream probes read the body.

Also minor: the PR description lists a `DEFAULT_GRACE=120.0` constant under "Constants mirror mech," but it isn't in the diff — likely dropped in the nits commit. Tidy the description for accuracy.

LOCKhart07 (approved):

OjusWiZard (approved):
Approving. The core fix is correct and well-tested: a frozen FSM now reports `is_healthy: false` and the `is_tm_healthy` None-coercion bug is fixed. The follow-up commit cleanly addressed the dead-operand simplification, docstring framing, and guarded `period_count` behind the AbciApp check to preserve the always-HTTP-200 contract. CI is fully green across the matrix.

Two advisory nits remain open and are non-blocking (a `reset_pause <= 0` guard and a redundant `bool()` cast); optionally tighten the `tm-unhealthy`/`stuck-no-transition` tests to assert the new `period: null` contract. None block merge.

PR #228, opened 2026-06-24
feat: guard market-creation cycle on insufficient wxDAI funding

jmoreira-valory:
## Summary

Guard the market-creation cycle so the service does **not** enter the proposal -> Mech-request -> market-creation flow when the safe cannot fund even one market. Today, when wxDAI is insufficient, the `createFPMM` multisend reverts on-chain (**GS013**) and the **Mech request fee is wasted** every cycle.

Observed in production (QS / `market_creator_ci`): after the last restart, 31 distinct `question_id` build-attempts but **only 15 markets created on-chain** -- the rest reverting `GS013`, plus ~80 Mech request/response cycles burned. The Pearl instance (better funded) created 18/18 cleanly.

## Behaviour

At `CollectProposedMarketsRound` (the entry to the cycle, reached after the wxDAI deposit has settled so the balance is final), read the safe's wxDAI balance via `ERC20.check_balance` and compare to the per-market requirement -- `to_wei(initial_funds / 100)`, the amount `createFPMM` pulls. If short:

- log an explicit error (`Insufficient wxDAI to create a market: have X, need Y ...`), and
- emit a new `Event.INSUFFICIENT_FUNDS` -> `FinishedWithoutTxRound` (the existing "nothing to do -> reset/pause" exit),

skipping the Mech request, processing, retrieval and creation entirely. On a transient balance-read failure the cycle is **not** blocked (fail-open).

## FSM change (synced across all 4 places)

`CollectProposedMarketsRound`: `INSUFFICIENT_FUNDS -> FinishedWithoutTxRound`
- `states/base.py` `Event` enum
- `rounds.py` transition_function + class docstring (regenerated)
- `fsm_specification.yaml` for **both** `market_creation_manager_abci` and the composed `market_maker_abci` (regenerated via `autonomy analyse fsm-specs --update`)

## Tests

- round `end_block`: `INSUFFICIENT_FUNDS_PAYLOAD -> Event.INSUFFICIENT_FUNDS` + the constant
- behaviour: `_have_funds_for_market` sufficient / insufficient (+ error log) / read-failure, and the `async_act` insufficient-funds branch
- **100% statement+branch coverage retained** on both changed files; full package suite (275 tests) green

## Scope / follow-ups

- Covers the **wxDAI (createFPMM) funding** -- the documented GS013 cause. It does not separately gate the native-xDAI Realitio bounty or gas (possible follow-up).
- Uses the same `to_wei(initial_funds / 100)` formula the factory contract applies.


Adamantios (dismissed):

packages/valory/skills/market_creation_manager_abci/behaviours/collect_proposed_markets.py
```diff
@@ -202,6 +221,34 @@ def async_act(self) -> Generator:
...
+            contract_address=self.params.collateral_tokens_contract,
+            contract_id=str(ERC20TokenContract.contract_id),
+            contract_callable="check_balance",
+            account=self.synchronized_data.safe_contract_address,
+            chain_id=self.params.default_chain_id,
+        )
+        if response.performative != ContractApiMessage.Performative.STATE:
+            self.context.logger.warning(f"check_balance unsuccessful!: {response}")
+            return True
+        balance = cast(int, response.state.body["token"])
```
OjusWiZard: **[HIGH] Unguarded `KeyError` silently bypasses the guard**

`response.state.body["token"]` raises if the key is absent (e.g. a schema change in the ERC20 wrapper or a malformed response). The performative check above only guards against non-`STATE` responses. On a `KeyError` the generator crashes, the round expires via `ROUND_TIMEOUT → RetrieveApprovedMarketRound`, and market creation proceeds without this guard ever firing — the exact failure mode the PR prevents.

```python
balance = response.state.body.get("token")
if balance is None:
    self.context.logger.warning("check_balance STATE response missing 'token' key")
    return True
balance = int(balance)
```

Also needs a test: `resp.state.body = {}` with `STATE` performative — `MagicMock` subscript access returns a mock rather than raising, so the crash is invisible to the current test suite.

packages/valory/skills/market_creation_manager_abci/behaviours/collect_proposed_markets.py
```diff
@@ -202,6 +221,34 @@ def async_act(self) -> Generator:
 
         self.set_done()
 
+    def _have_funds_for_market(self) -> Generator[None, None, bool]:
+        """Return True if the safe holds enough wxDAI to fund one market."""
+        # Compare the safe's wxDAI (collateral) balance against the amount
+        # ``createFPMM`` pulls -- ``to_wei(initial_funds / 100)``. On a transient
+        # balance-read failure the cycle is not blocked (returns True).
+        required = int(self.params.initial_funds / 100 * 10**18)
```
OjusWiZard: **[MEDIUM] Float arithmetic may diverge from the factory contract's `to_wei`**

`int(self.params.initial_funds / 100 * 10**18)` uses IEEE-754 float multiplication. The factory contract computes the actual on-chain spend via `web3.to_wei(initial_funds / math.pow(10, 2), "ether")`, which uses `Decimal` internally. For non-round values the two paths differ by up to 128 wei:

| `initial_funds` | Guard threshold | `to_wei` result | Effect |
|---|---|---|---|
| `100.1` | `1_000_999_999_999_999_872` | `1_001_000_000_000_000_000` | Guard 128 wei **low** → false negative (guard passes, on-chain reverts) |
| `99.9` | `999_000_000_000_000_128` | `999_000_000_000_000_000` | Guard 128 wei high → spurious skip |

Fix — exact integer equivalent of `to_wei(x/100, "ether")`:
```python
required = int(self.params.initial_funds * 10**16)
```

packages/valory/skills/market_creation_manager_abci/behaviours/collect_proposed_markets.py
```diff
@@ -71,9 +73,26 @@ class CollectProposedMarketsBehaviour(MarketCreationManagerBaseBehaviour):
...
     def async_act(self) -> Generator:
         """Do the act, supporting asynchronous execution."""
+        sender = self.context.agent_address
+
+        # Guard: skip the entire propose -> Mech-request -> create cycle when
+        # the safe cannot fund even one market (initial_funds). Creating it
+        # would only revert on-chain (GS013) and waste the Mech request fee, so
+        # branch out via INSUFFICIENT_FUNDS and let the period reset/pause.
+        with self.context.benchmark_tool.measure(self.behaviour_id).local():
+            have_funds = yield from self._have_funds_for_market()
```
OjusWiZard: **[MEDIUM] Two consecutive `.local()` blocks on the happy path**

The fund check opens and closes a `.local()` block here (lines 82–83), then the happy path opens a second `.local()` block at line 95. Every other behaviour in this skill uses exactly one `.local()` for all local work. The second measurement silently overwrites or drops the first in the benchmark framework, so the contract API round-trip time is lost from metrics.

Consider moving the `_have_funds_for_market` call inside the single `.local()` block that already wraps the main local computation.

packages/valory/skills/market_creation_manager_abci/tests/behaviours/test_collect_proposed_markets.py
```diff
@@ -190,6 +191,49 @@ def test_matching_round(self) -> None:
...
+        resp = MagicMock()
+        resp.performative = ContractApiMessage.Performative.STATE
+        resp.state.body = {"token": 5 * 10**17}  # 0.5 wxDAI < 1 wxDAI
+        with patch.object(
+            self.behaviour, "get_contract_api_response", new=_make_gen(resp)
+        ):
+            result = _exhaust_gen(self.behaviour._have_funds_for_market())
+        assert result is False
+        errors = [str(c) for c in self.behaviour.context.logger.error.call_args_list]
+        assert any("Insufficient wxDAI" in c for c in errors)
```
OjusWiZard, line 222: **[TEST] Missing edge case: `STATE` response with absent `"token"` key**

The three `_have_funds_for_market` tests cover `STATE`+key-present (sufficient), `STATE`+key-present (insufficient), and `ERROR` performative. There is no test for `STATE` + `body = {}`. Because `MagicMock` subscript access returns another mock rather than raising `KeyError`, the real crash is invisible to the current suite — coverage passes but the production path is untested.

Add:
```python
def test_have_funds_for_market_missing_token_key(self) -> None:
    resp = MagicMock()
    resp.performative = ContractApiMessage.Performative.STATE
    resp.state.body = {}  # "token" key absent
    with patch.object(self.behaviour, "get_contract_api_response", new=_make_gen(resp)):
        result = _exhaust_gen(self.behaviour._have_funds_for_market())
    assert result is True  # fail-open
```

packages/valory/skills/market_creation_manager_abci/tests/behaviours/test_collect_proposed_markets.py
```diff
@@ -414,13 +462,39 @@ def test_async_act_unprocessed_markets_exist(self) -> None:
...
+            patch.object(self.behaviour, "send_a2a_transaction", new=_capture_send),
+            patch.object(self.behaviour, "wait_until_round_end", new=_make_gen(None)),
+            patch.object(self.behaviour, "set_done") as mock_set_done,
+        ):
+            gen = self.behaviour.async_act()
+            _exhaust_gen(gen)
+            mock_set_done.assert_called_once()
+            assert (
+                captured["payload"].content
+                == CollectProposedMarketsRound.INSUFFICIENT_FUNDS_PAYLOAD
```
OjusWiZard: **[TEST] `wait_until_round_end` is not asserted**

The test verifies `set_done()` and `payload.content` but does not assert that `wait_until_round_end` was called. That call is what blocks the behaviour until the round transitions — if it were accidentally dropped from the source, this test would still pass while the FSM stalls.

packages/valory/skills/market_creation_manager_abci/tests/behaviours/test_collect_proposed_markets.py
```diff
@@ -414,13 +462,39 @@ def test_async_act_unprocessed_markets_exist(self) -> None:
...
             patch.object(self.behaviour, "set_done"),
         ):
             gen = self.behaviour.async_act()
             _exhaust_gen(gen)
 
+    def test_async_act_insufficient_funds(self) -> None:
+        """async_act branches to INSUFFICIENT_FUNDS_PAYLOAD when funds are short."""
+        captured = {}
+
+        def _capture_send(payload: Any, *a: Any, **k: Any) -> Any:
```
OjusWiZard: **[LOW] Unreachable-yield generator idiom**

`return; yield` works but suppresses a linter warning for a non-obvious reason. `_make_gen` already exists in this file for exactly this use case — prefer it and capture the call via `mock.call_args` instead of a custom closure.

packages/valory/skills/market_creation_manager_abci/behaviours/collect_proposed_markets.py
```diff
@@ -202,6 +221,34 @@ def async_act(self) -> Generator:
...
+        # Compare the safe's wxDAI (collateral) balance against the amount
+        # ``createFPMM`` pulls -- ``to_wei(initial_funds / 100)``. On a transient
+        # balance-read failure the cycle is not blocked (returns True).
+        required = int(self.params.initial_funds / 100 * 10**18)
+        response = yield from self.get_contract_api_response(
+            performative=ContractApiMessage.Performative.GET_STATE,  # type: ignore
+            contract_address=self.params.collateral_tokens_contract,
+            contract_id=str(ERC20TokenContract.contract_id),
+            contract_callable="check_balance",
+            account=self.synchronized_data.safe_contract_address,
```
OjusWiZard, line 221: **[IMPORTANT] Unguarded `ValueError` on `int(balance)` for non-integer token values**

The docstring states "On a transient balance-read failure or a missing `token` key the cycle is not blocked" — but a `STATE` response with a non-numeric `"token"` value (e.g. hex string `"0xDE0B6B3A7640000"`) raises an unhandled `ValueError` here. That exception propagates out of `async_act`, crashing the behaviour and causing `ROUND_TIMEOUT → RetrieveApprovedMarketRound` instead of the intended fail-open `True`. The framework's ABI decoder returns Python `int` for `uint256` so this is low-probability in practice, but the fail-open intent is not enforced for this code path and there is no test covering it.

Consider wrapping with a guard consistent with the existing pattern:
```python
try:
    balance = int(balance)
except (ValueError, TypeError):
    self.context.logger.warning(f"check_balance returned non-integer token value: {balance!r}")
    return True
```

OjusWiZard (approved):
All prior findings addressed (KeyError guard, float precision, double `.local()`, missing-key test, `wait_until_round_end` assertion, `return;yield` idiom). One remaining minor point left inline on `int(balance)` — the fail-open intent in the docstring is not enforced for a non-numeric token value. Low probability given the framework ABI decoder, but worth a try/except guard for consistency. LGTM otherwise.

Adamantios (approved):

PR #230, opened 2026-06-26
chore(deps): bump omen_realitio_withdraw_bonds_abci CID (resume-after-timeout cache)

jmoreira-valory:
## Summary

Cascades the resume-after-timeout claim cache from [valory-xyz/omen-protocol#3](https://github.com/valory-xyz/omen-protocol/pull/3) (tag `v0.1.2`) into market-creator, for consistency with the market-resolver bump ([valory-xyz/market-resolver#34](https://github.com/valory-xyz/market-resolver/pull/34)).

\`\`\`
third_party CID
- bafybeicj3pgdx7q5chbms3pxrb6pcv7lbqhkigwnmjuo54p63ub7mkqe5i
+ bafybeiabqn6mcg4nw3mqbyxcuiyxgwlf7e2il3cj4awpmdi7bibqlwvklq
\`\`\`

Plus the omen-protocol upstream pin bump (`v0.1.1` → `v0.1.2`) so `check-third-party-hashes` recognises the new CID, and the four cascaded dev hashes (`market_creation_manager_abci`, `market_maker_abci`, agent, service). No dev source touched.

## Why market-creator needs it too

`market_maker_abci/composition.py` wires the `omen_realitio_withdraw_bonds_abci` round into the creator's FSM (confirmed in `composition.py`, `skill.yaml`, `behaviours.py`, `models.py`). So market-creator runs the same bond-recovery flow and carries the same starvation risk that was fixed: under a 60s `round_timeout`, the RPC-heavy multisend build (`eth_getLogs` + `eth_call` simulation + safe-tx build, per claim) can run past the deadline and be discarded, restarting from scratch every cycle and never draining the claimable queue.

## What the upstream change does

Caches per-question claim artifacts on `SharedState.realitio_claim_build_cache`. A round that times out before settling reuses the build next cycle instead of rebuilding. Each cache hit re-runs `_simulate_claim` to detect calldata that would now revert (subgraph lagging behind a settled claim), evicting stale entries — protecting `MultiSend.delegatecall requireSuccess=true` atomicity.

## Validation

The fix was smoke-tested against a Tenderly Gnosis fork (real `RealitioContract` + skill code): a synthetic question created on the fork, answered, time-advanced to finalize, then the **real `_build_claim_txs` cache loop** driven through cold-build → warm-reuse (rebuild skipped) → post-settlement stale-eviction, with on-chain `claimWinnings` simulation correctly flipping True→False. All paths behaved as designed.

## Test plan

[...]

PR #231, opened 2026-06-30
fix(models): wire RealitioWithdrawBonds SharedState into composed MRO (crash-loop fix)

jmoreira-valory:
## Same crash-loop fix as market-resolver#35, applied to market-creator

After the resume-after-timeout claim cache (`omen_realitio_withdraw_bonds_abci` v0.1.2, cascaded via #230) merged, market-creator `main` was left in the same crash-prone state the resolver hit in production: the composed `market_maker_abci.SharedState` does not include the withdraw-bonds sub-skill in its MRO, so `realitio_claim_build_cache` is never initialised on the runtime `self.context.state`.

`RealitioWithdrawBondsBehaviour._build_claim_txs` reads it directly:

\`\`\`
AttributeError: 'SharedState' object has no attribute 'realitio_claim_build_cache'
-> AEAActException -> skill_exception_policy: stop_and_exit -> Propel restart loop
\`\`\`

The market-creator deployment was **not** yet bumped to the cache CID (the agent-deployments env bump was held), so this is a pre-deploy fix — no production impact, unlike the resolver.

## Fix — standard repo pattern

The composed `SharedState` already multiply-inherits `MechInteractSharedState` and `CtRedeemTokensSharedState` so their `__init__` cross-round vars (`ignored_ct_positions`, mech state) land on the runtime state. This adds `RealitioWithdrawBondsSharedState` alongside them:

\`\`\`python
class SharedState(
    MechInteractSharedState,
    BaseSharedState,
    CtRedeemTokensSharedState,
    RealitioWithdrawBondsSharedState,   # <- added
):
\`\`\`

The class docstring is updated to spell out the invariant (omitting a sub-skill whose behaviour reads `self.context.state.<attr>` crash-loops).

## Verification
\`\`\`
instantiate composed SharedState -> realitio_claim_build_cache == {}  (next to ignored_ct_positions)
\`\`\`

## Test plan
- [x] `TestSharedState` (4) incl. new regression (attr initialised + sub-skill in MRO)
- [x] `market_maker_abci` suite **58 passed**
- [x] lint green (isort/black/flake8/mypy/pylint/darglint)
- [x] structural green (check-abciapp-specs/handlers/packages)
- [x] `autonomy packages lock` cascade; `third_party` preserved

## Rollout
Merge → service CID `bafybeibb5msj3gggv2clat37k4vfpalsjr6bveym3ve6ty64ean3tpgpzy`. The held agent-deployments creator env bumps (QS + Pearl) must target this fixed CID, not #230's.

### DIFF
--- a/packages/valory/skills/market_maker_abci/tests/test_models.py
+++ b/packages/valory/skills/market_maker_abci/tests/test_models.py
@@ -86,6 +86,27 @@ class TestSharedState:
         """Test inheritance."""
         assert issubclass(SharedState, BaseSharedState)
 
+    def test_realitio_claim_build_cache_initialized(self) -> None:
+        """The withdraw-bonds claim cache must be initialized on the shared state.
+
+        Regression: ``RealitioWithdrawBondsBehaviour._build_claim_txs``
+        reads ``self.context.state.realitio_claim_build_cache``. That
+        attribute only exists if ``RealitioWithdrawBondsSharedState`` is in
+        this composed ``SharedState``'s MRO so its ``__init__`` runs via the
+        cooperative ``super()`` chain (same mechanism as
+        ``ignored_ct_positions``). When it was absent the round raised
+        ``AttributeError`` -> ``stop_and_exit`` -> Propel restart loop.
+        """
+        from unittest.mock import MagicMock
+
+        from packages.valory.skills.omen_realitio_withdraw_bonds_abci.models import (
+            SharedState as RealitioWithdrawBondsSharedState,
+        )
+
+        assert RealitioWithdrawBondsSharedState in SharedState.__mro__
+        state = SharedState(name="state", skill_context=MagicMock())
+        assert state.realitio_claim_build_cache == {}
+
     def test_abci_app_cls(self) -> None:
         """Test abci_app_cls is set."""
         from packages.valory.skills.market_maker_abci.composition import (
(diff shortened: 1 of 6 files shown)


packages/valory/skills/market_maker_abci/tests/test_models.py
```diff
@@ -86,6 +86,27 @@ def test_inherits_base_shared_state(self) -> None:
...
+        this composed ``SharedState``'s MRO so its ``__init__`` runs via the
+        cooperative ``super()`` chain (same mechanism as
+        ``ignored_ct_positions``). When it was absent the round raised
+        ``AttributeError`` -> ``stop_and_exit`` -> Propel restart loop.
+        """
+        from unittest.mock import MagicMock
+
+        from packages.valory.skills.omen_realitio_withdraw_bonds_abci.models import (
+            SharedState as RealitioWithdrawBondsSharedState,
+        )
```
Adamantios, line 104: Why local imports?

Adamantios (approved):

OjusWiZard (approved):
Approve — minimal, well-targeted crash-loop fix.

- MRO is sound: `RealitioWithdrawBondsSharedState` is appended last, so it can't override any previously-resolved method; its only effect is contributing `__init__` to the cooperative `super()` chain (same mechanism as `ignored_ct_positions`).
- Regression test asserts both MRO membership and `realitio_claim_build_cache == {}` after real instantiation — the right pair of assertions for this bug class.
- Import follows the file's existing force_single_line isort convention.
- Hash cascade is internally consistent (skill CID matches aea-config; agent CID matches service.yaml; service CID matches the rollout CID in the PR body).
- Docstring now documents the invariant.

LGTM.

packages/valory/skills/market_maker_abci/tests/test_models.py
jmoreira-valory: Good catch — no real reason. The `MagicMock` import here is redundant (already imported module-level at the top of the file), and `RealitioWithdrawBondsSharedState` could be hoisted to module level too (no circular-import concern; `market_maker_abci.models` already pulls it in). Leaving it as-is on this PR to avoid bouncing the approval for a cosmetic change, but I'll clean it up in a follow-up touch to this file. Thanks!

DIvyaNautiyal07 (approved):

